PR #3, opened 2025-01-10
Initial redesign

TheOneRing:


THEME.cmake
```diff
@@ -41,14 +41,6 @@ string(REPLACE "&" "&amp;" APPLICATION_NAME_XML_ESCAPED "${APPLICATION_NAME}")
...
 string(REPLACE ">" "&gt;" APPLICATION_NAME_XML_ESCAPED "${APPLICATION_NAME_XML_ESCAPED}")
 
-if (NOT DEFINED LINUX_PACKAGE_SHORTNAME)
-    set(LINUX_PACKAGE_SHORTNAME "${APPLICATION_SHORTNAME}")
-endif()
-
-if (NOT DEFINED PACKAGE)
-    set(PACKAGE "${LINUX_PACKAGE_SHORTNAME}-client")
-endif()
-
```
dragotin, line 51: Why is this removed?

src/gui/generalsettings.ui
```diff
@@ -230,19 +230,6 @@
...
-         <cursorShape>ArrowCursor</cursorShape>
-        </property>
-        <property name="text">
-         <string notr="true">Copyright ownCloud GmbH</string>
-        </property>
-        <property name="textInteractionFlags">
-         <set>Qt::LinksAccessibleByMouse|Qt::TextSelectableByMouse</set>
-        </property>
-       </widget>
-      </item>
```
dragotin, line 245: yes! We could add the whole "About" stuff underneath the current content in the settings tab and save the button, but that is for later.

src/libsync/theme.cpp
```diff
@@ -243,7 +243,11 @@ void Theme::setSystrayUseMonoIcons(bool mono)
 
 QUrl Theme::updateCheckUrl() const
 {
+#ifdef APPLICATION_UPDATE_URL
```
dragotin, line 246: isn't update**Check**Url != APP_**UPDATE**_URL ??

THEME.cmake
TheOneRing: The change was unrelated but the var is unused

src/gui/generalsettings.ui
TheOneRing: Its a dialogue, and we also display it from the system tray.

PR #13, opened 2025-01-14
Update about text

TheOneRing:


src/libsync/theme.cpp
```diff
@@ -317,22 +319,19 @@ QString Theme::aboutVersions(Theme::VersionFormat format) const
 
 QString Theme::about() const
 {
-    // Ideally, the vendor should be "ownCloud GmbH", but it cannot be changed without
-    // changing the location of the settings and other registery keys.
-    const QString vendor = Resources::isVanillaTheme() ? QStringLiteral("ownCloud GmbH") : QStringLiteral(APPLICATION_VENDOR);
     return tr("<p>Version %1. For more information visit <a href=\"%2\">https://%3</a></p>"
-              "<p>For known issues and help, please visit: <a href=\"https://central.owncloud.com/c/desktop-client\">https://central.owncloud.com</a></p>"
+              "<p>For known issues and help, please visit: <a href=\"https://github.com/opencloud-eu/desktop\"></p>"
```
dragotin: 1. Does this display the link? Note that the original text had a link text as well
2. Github is different from the community discussion board which was central. I think we need to change that back to a less developer oriented board once we have it.

src/libsync/theme.cpp
```diff
@@ -287,12 +287,14 @@ QString Theme::aboutVersions(Theme::VersionFormat format) const
         Q_UNREACHABLE();
     }();
     const QString qtVersion = QString::fromUtf8(qVersion());
-    const QString qtVersionString = (QLatin1String(QT_VERSION_STR) == qtVersion ? qtVersion : QCoreApplication::translate("ownCloudTheme::qtVer", "%1 (Built against Qt %2)").arg(qtVersion, QStringLiteral(QT_VERSION_STR)));
+    const QString qtVersionString = (QLatin1String(QT_VERSION_STR) == qtVersion
+            ? qtVersion
+            : QCoreApplication::translate("OpenCloudTheme::qtVer", "%1 (Built against Qt %2)").arg(qtVersion, QStringLiteral(QT_VERSION_STR)));
```
dragotin, line 296: Have you dropped support for name theming already as the _OpenCloud_ is hardcoded here?

src/libsync/theme.cpp
```diff
@@ -317,22 +319,19 @@ QString Theme::aboutVersions(Theme::VersionFormat format) const
...
               "<p><small>By Klaas Freitag, Daniel Molkentin, Olivier Goffart, Markus Götz, "
               " Jan-Christoph Borchardt, Thomas Müller,<br>"
               "Dominik Schmidt, Michael Stingl, Hannah von Reth, Fabian Müller and others.</small></p>"
-              "<p>Copyright ownCloud GmbH</p>"
-              "<p>Distributed by %4 and licensed under the GNU General Public License (GPL) Version 2.0.<br/>"
-              "%5 and the %5 logo are registered trademarks of %4 in the "
+              "<p>Copyright 2014-2024 ownCloud GmbH</p>"
+              "<p>Copyright OpenCloud GmbH</p>"
+              "<p>Distributed by OpenCloud GmbH and licensed under the GNU General Public License (GPL) Version 2.0.<br/>"
+              "%5 and the %4 logo are registered trademarks of OpenCloud GmbH in the "
```
dragotin: Puh... I am not sure if we still need the "United States and other countries" foo. I will check with the lawyer.

src/libsync/theme.cpp
```diff
@@ -317,22 +319,19 @@ QString Theme::aboutVersions(Theme::VersionFormat format) const
...
               "United States, other countries, or both.</p>"
-              "<p><small>%6</small></p>")
+              "<p><small>%5</small></p>")
         .arg(Utility::escape(Version::displayString()), Utility::escape(QStringLiteral("https://" APPLICATION_DOMAIN)),
-            Utility::escape(QStringLiteral(APPLICATION_DOMAIN)), Utility::escape(vendor), Utility::escape(appNameGUI()),
-            aboutVersions(Theme::VersionFormat::RichText));
+            Utility::escape(QStringLiteral(APPLICATION_DOMAIN)), Utility::escape(appNameGUI()), aboutVersions(Theme::VersionFormat::RichText));
 }
```
dragotin, line 332: The new code looks different from the params... - is that intentional?

src/libsync/theme.cpp
TheOneRing: Yes, the change in 322 removes the support of white label builds.

src/libsync/theme.cpp
TheOneRing: Yes the texts was missing and yes once we have a better place to point the users...

src/libsync/theme.cpp
TheOneRing: Me neither but its the original ownCloud text...

src/libsync/theme.cpp
TheOneRing: it removed the vendor which was used in
`logo are registered trademarks of %4 in the`

PR #35, opened 2025-01-21
Use OpenCloud fonts

TheOneRing:
image
image
image

image


TheOneRing:
Hmm looks like the GroupBox title is not using the correct font.

PR #53, opened 2025-01-28
Minor cleanup after removal of chunking v1

TheOneRing:


src/libsync/propagateuploadv1.cpp
```diff
@@ -57,32 +55,19 @@ void PropagateUploadFileV1::doStartUpload()
...
-
     if (!_transmissionChecksumHeader.isEmpty()) {
         qCInfo(lcPropagateUploadV1) << propagator()->fullRemotePath(path) << _transmissionChecksumHeader;
         headers[checkSumHeaderC] = _transmissionChecksumHeader;
     }
 
-    const QString fileName = propagator()->fullLocalPath(_item->localName());
-    auto device = std::make_unique<UploadDevice>(fileName, chunkStart, currentChunkSize,
-        propagator()->_bandwidthManager);
+    auto device = std::make_unique<UploadDevice>(fileName, 0, fileSize, propagator()->_bandwidthManager);
```
dragotin, line 70: Not sure where "fileName" comes from... but that's probably outside the diff.

PR #56, opened 2025-01-29
chore: use main instead of master for craft related files

kulmann:
`craft` related files were referencing `master` as the main branch of this repo. fixed that.

### DIFF
--- a/.craft.ini
+++ b/.craft.ini
@@ -1,5 +1,4 @@
 [General]
-Branch = master
 ShallowClone = False
 
 # Variables defined here override the default value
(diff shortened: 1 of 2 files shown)


.craft.ini
```diff
@@ -1,5 +1,5 @@
 [General]
-Branch = master
+Branch = main
```
TheOneRing: Hm that's a config flag for https://invent.kde.org/packaging/craftmaster and I guess we can remove it

kulmann: If you say so... I removed it 😅

PR #67, opened 2025-02-03
Announce to the user that a sync will happen, as soon as we detect a change

TheOneRing:
@JammingBen

### DIFF
--- a/src/gui/folderwatcher.h
+++ b/src/gui/folderwatcher.h
@@ -18,10 +18,8 @@
 
 
 #include <QElapsedTimer>
-#include <QHash>
 #include <QList>
 #include <QLoggingCategory>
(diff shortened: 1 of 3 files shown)


src/gui/folderwatcher.h
```diff
@@ -91,6 +89,12 @@ class OPENCLOUD_GUI_EXPORT FolderWatcher : public QObject
      *  of the contained files is changed. */
     void pathChanged(const QSet<QString> &path);
 
+    /**
+     * We detected a file change, this signal can be used to trigger the prepareSync state
+     */
+    // TODO: rename
```
TheOneRing, line 95: Naming things is hard, but we already have a function `changeDetected`

PR #141, opened 2025-03-12
Fix the scroll view

TheOneRing:


src/gui/spaces/qml/SpacesView.qml
```diff
@@ -31,11 +31,12 @@ Pane {
     Accessible.name: qsTr("Spaces")
 
     spacing: 10
-
     ScrollView {
         id: scrollView
         anchors.fill: parent
         clip: true
+        rightPadding: spacesView.spacing * 2
```
TheOneRing, line 38: I really can't tell what is different compared with the scroll view in https://github.com/opencloud-eu/desktop/pull/141/commits/259551dbbd7e84da6e867bb6d6a67bf2e3f6fbe2#diff-71635d4152ed7d04fc588c057b9d07710bb4bbad353bf53b5f450e684825cd4bR147

However, just setting padding here would result in a padding on all sides BUT the scrollbar side.
Setting 1*spacing also has not visible effect, and only multiplied by 2 has a visible effect.

PR #149, opened 2025-03-17
Add dev profile

TheOneRing:


OPENCLOUD.cmake
```diff
@@ -1,8 +1,15 @@
-set( APPLICATION_NAME       "OpenCloud Desktop" )
-set( APPLICATION_SHORTNAME  "OpenCloud" )
-set( APPLICATION_EXECUTABLE "opencloud" )
-set( APPLICATION_DOMAIN     "opencloud.eu" )
-set( APPLICATION_VENDOR     "OpenCloud" )
+option(CANARY_BUILD "Use a standalone profile" OFF)
+
+if (CANARY_BUILD)
+    set(THEME_SUFFIX " Canary")
```
dragotin: Can we use a different term than "Canary"? Something like "Head" or "git" for example?

src/resources/resources.cpp
```diff
@@ -58,7 +58,7 @@ QString brandThemePath()
 
 bool Resources::isVanillaTheme()
 {
-    return std::string_view(APPLICATION_SHORTNAME) == "OpenCloud";
+    return std::string_view(APPLICATION_SHORTNAME).starts_with("OpenCloud");
```
dragotin, line 61: This seems to cry for trouble as we move forward... Why not set a flag in CMake?

dragotin (changes requested):

PR #176, opened 2025-04-08
Fix dynamic registration with public client

TheOneRing:
With a public client we won't receive a client secret and no expiration date for that client secret.

### DIFF
--- a/src/libsync/creds/oauth.cpp
+++ b/src/libsync/creds/oauth.cpp
@@ -461,6 +461,16 @@ QUrl OAuth::authorisationLink() const
     return Utility::concatUrlPath(_authEndpoint, {}, query);
 }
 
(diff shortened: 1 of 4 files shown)


rhafer:
> With a public client we won't receive a client secret and no expiration date for that client secret.

I don't really know, but do we really always only want to support public clients for dynamic client registration?

Our default client is public, because we can't keep the secret confidential anyway. But when dynamically registering we might benefit if we allow registereing a confidential client, no?

micbar:
> Our default client is public, because we can't keep the secret confidential anyway. But when dynamically registering we might benefit if we allow registereing a confidential client, no?

IMHO we should at least have the option to choose.

Registering a not public client is a huge security gain, because the client id and secret are not shared with others.

PR #182, opened 2025-04-10
Correctly set pointer exposed to qml to null

TheOneRing:


src/gui/settingsdialog.cpp
```diff
@@ -242,9 +245,13 @@ SettingsDialog::SettingsPage SettingsDialog::currentPage() const
...
     _currentAccount = account;
-    _ui->stack->setCurrentWidget(accountSettings(account));
-    _currentPage = SettingsPage::Account;
-
+    if (account) {
+        _ui->stack->setCurrentWidget(accountSettings(account));
+        _currentPage = SettingsPage::Account;
+    } else {
+        // go to the settings page if the last account was removed
+        _ui->stack->setCurrentWidget(_generalSettings);
```
dragotin: In this case, `_currentPage` is not changed at all. No idea if that is a problem or not...

dragotin (approved):

src/gui/settingsdialog.cpp
TheOneRing: your right thx

src/gui/settingsdialog.cpp
```diff
@@ -242,11 +245,16 @@ SettingsDialog::SettingsPage SettingsDialog::currentPage() const
 void SettingsDialog::setCurrentAccount(Account *account)
 {
     _currentAccount = account;
-    _ui->stack->setCurrentWidget(accountSettings(account));
-    _currentPage = SettingsPage::Account;
+    if (account) {
+        _ui->stack->setCurrentWidget(accountSettings(account));
+        _currentPage = SettingsPage::Account;
```
dragotin, line 250: Ok, now shouldn't this also be `setCurrentPage(SettingsPage::Account);`? Similar to the else branch?

TheOneRing: Hmm not in the current design of that class.

TheOneRing: https://github.com/opencloud-eu/desktop/pull/182/files#diff-62aa619164fed2fe67a05f08aab80e5d2dff11db741e69a2ce68981fb1bcb37bR229

Issue #222, opened 2025-04-22
Qt 6.8.3 or later. Re-enable native tooltips


TheOneRing:
We suspect a bug to be present in Qt 6.8.1 that causes crashes related to the native popups in the account bar.
On Wayland, the application becomes unresponsive and on X11 we sometimes get a crash.

TheOneRing:
x11 crash:
```
1  QWindow::mapFromGlobal                                                                                                                          qwindow.cpp                2969 0x7ffff67ff5b6 
2  QWindow::mapFromGlobal                                                                                                                          qwindow.cpp                2998 0x7ffff67ff78c 
3  QQuickPopupWindow::global2Local                                                                                                                 qquickpopupwindow.cpp      343  0x7fffd7128220 
4  QQuickPopupWindow::resizeEvent                                                                                                                  qquickpopupwindow.cpp      122  0x7fffd712966d 
5  QWindow::event                                                                                                                                  qwindow.cpp                2610 0x7ffff6802cd8 
6  QQuickWindow::event                                                                                                                             qquickwindow.cpp           1641 0x7ffff44aa767 
7  QQuickWindowQmlImpl::event                                                                                                                      qquickwindowmodule.cpp     141  0x7ffff44b9a96 
8  QQuickPopupWindow::event                                                                                                                        qquickpopupwindow.cpp      327  0x7fffd7129838 
9  QApplicationPrivate::notify_helper                                                                                                              qapplication.cpp           3296 0x7ffff7187fe7 
10 QApplication::notify                                                                                                                            qapplication.cpp           3247 0x7ffff718eedc 
11 QCoreApplication::notifyInternal2                                                                                                               qcoreapplication.cpp       1168 0x7ffff5753bd6 
12 QCoreApplication::sendSpontaneousEvent                                                                                                          qcoreapplication.cpp       1626 0x7ffff5753cd3 
13 QGuiApplicationPrivate::processGeometryChangeEvent                                                                                              qguiapplication.cpp        2852 0x7ffff67a4205 
14 QGuiApplicationPrivate::processWindowSystemEvent                                                                                                qguiapplication.cpp        2200 0x7ffff67a791d 
15 QWindowSystemInterface::sendWindowSystemEvents                                                                                                  qwindowsysteminterface.cpp 1114 0x7ffff6805c9c 
16 xcbSourceDispatch                                                                                                                               qxcbeventdispatcher.cpp    57   0x7ffff2f8c928 
[...]

Issue #212, opened 2025-04-17
Afte adding a second account to a server running on localhost the client hangs


individual-it:
After I've added a second account to a server running on my development machine (stated with the full deployment example plus `STORAGE_USERS_POSIX_WATCH_FS: true`) and playing around the client started to hang, till I removed that account

Logs can be found in https://opencloud.in-nepal.de/s/YLiAQKYUOprWVNN (password will be shared privately as the logs contain also tokens to a server in production)

Issue #207, opened 2025-04-16
Crash after configuring account


dragotin:
Using 1.0.0 RC3 AppImage on Linux with Server version 2.1.0 deployed on localhost:9200 via simple start script.

After configuring the account in the client with all standards, the client crashes when pressing "ok".

This is the (useless) output on the terminal:
```bash
[kf:~] $ ./OpenCloud_Desktop-v1.0.0-rc.3-linux-gcc-x86_64.AppImage 
kf.kio.gui: Not passing environment variable "BASH_FUNC_mc%%" to systemd because its name contains illegal characters
Speicherzugriffsfehler (Speicherabzug geschrieben)
```

TheOneRing:
Please reproduce the crash with logs enabled, we won't get a backtrace, but the logs might tell us what happened.

Issue #195, opened 2025-04-14
Desktop client crashes when switching from activitylog to account overview


dragonchaser:
When you switch from the activitylog (independent of the selected tab) back to the account overview the client will often become unresponsive and crash.

**System:** 
- arch
- plasmadesktop
- opencloud desktop client (1.0.0 beta and rc.1)

PR #225, opened 2025-04-22
Disable native tooltips on all platforms due to macOS crashes

### LINKED ISSUE #224: Desktop App Crash/Hang on macOS related to Qt native popups
## Description
The OpenCloud Desktop application crashes or becomes unresponsive on macOS when displaying popup components, similar to the Linux issues described in #222.

## Platform
- macOS 15.4.1 (macOS Sequoia)
- OpenCloud Desktop 1.0.0-rc.3
- Architecture: arm64 (Apple Silicon)

## Crash Details
The crash report shows a hang in the main thread during `QQuickPopup::setVisible`, specifically in the call chain:

```
QQuickPopup::setVisible(bool) + 140
QQuickPopupPrivate::prepareEnterTransition() + 120
QQuickPopupPrivate::adjustPopupItemParentAndWindow() + 676
QQuickWindowQmlImpl::applyWindowVisibility() + 1032
```

This is happening even though native tooltips were only disabled for Linux platforms in commit bcd33c9c.

## Related Issues
- #222: Qt 6.8.3 or later. Re-enable native tooltips on Linux
- #212, #207, #195: Crash issues that were fixed by disabling native tooltips on Linux

## Suggested Fix
Consider disabling native tooltips on all platforms until the Qt bug is resolved, not just on Linux:

```qml
// Change from:
if (Qt.platform.os !== "linux") {
    if ('popupType' in ToolTip.toolTip) {
        ToolTip.toolTip.popupType = Popup.Native;
    }
}

// To:
if (false) { // Disable on all platforms until fixed
    if ('popupType' in ToolTip.toolTip) {
        ToolTip.toolTip.popupType = Popup.Native;
    }
}
```

michaelstingl:
## Summary
- Disable native tooltips on all platforms, not just Linux, as they also cause crashes on macOS
- This is a temporary workaround until the underlying Qt issue is fixed (expected in Qt 6.6.3 or later)

## Problem
As documented in issue #224, the OpenCloud Desktop application crashes or becomes unresponsive on macOS when hovering over account buttons. The crash log indicates a hang in the Qt tooltip/popup handling code similar to the issues previously observed on Linux platforms.

The crash occurs in:
```
QQuickPopup::setVisible(bool)
QQuickPopupPrivate::prepareEnterTransition()
QQuickPopupPrivate::adjustPopupItemParentAndWindow()
```

## Solution
This PR extends the existing workaround (which disabled native tooltips only on Linux platforms) to all platforms by changing the condition from:
```qml
if (Qt.platform.os !== "linux") {
    // Use native tooltips
}
```

to:
```qml
if (false) { // Disabled on all platforms instead of just Linux
    // Use native tooltips
}
```

This ensures that native tooltips are disabled on all platforms, preventing crashes on macOS as well as Linux.

## Related Issues
- Fixes #224 (macOS crashes)
- Related to #222 (Qt 6.8.3 or later: Re-enable native tooltips on Linux)
- Related to previous fixes for #212, #207, #195 (Linux crashes)

## Test Plan
- Hover mouse over account buttons or "Add Account" button in the desktop app
- Verify the app does not freeze or crash with spinning beach ball on macOS
- Ensure tooltips still appear but are rendered by QML instead of native tooltips

### DIFF
--- a/src/gui/qml/AccountBar.qml
+++ b/src/gui/qml/AccountBar.qml
@@ -29,7 +29,8 @@ Pane {
     Component.onCompleted: {
         // TODO: Re enable with Qt 6.6.3
         // https://github.com/opencloud-eu/desktop/issues/222
-        if (Qt.platform.os !== "linux") {
+        // Disabled on all platforms due to macOS crashes (issue #224)
+        if (false) { // Disabled on all platforms instead of just Linux
             if ('popupType' in ToolTip.toolTip) {
                 ToolTip.toolTip.popupType = Popup.Native;
             }


TheOneRing:
Hi Micheal,
thx for the patch, but I decided on removing the code completely, for now.
https://github.com/opencloud-eu/desktop/pull/228

Issue #226, opened 2025-04-22
Improve keyboard navigation in setup wizard


michaelstingl:
## Issue Description
Currently, the account setup wizard in the desktop app requires mouse interaction to proceed through each step. The Enter key doesn't work to confirm inputs or advance to the next step, and the Escape key doesn't close dialogs.

### Expected Behavior
- The Enter key should confirm/proceed at each step:
  - Server URL input: Enter should trigger the Continue button
  - Login screen: Enter should trigger the "Open web browser" button
  - Final screen: Enter should trigger the Done button
- The Escape key should cancel/close the wizard

### Current Behavior
- Enter key does nothing in most screens
- Escape key doesn't close the dialog
- User is forced to use the mouse for all interactions

### Impact
These keyboard shortcuts would significantly improve the usability of the application, especially for keyboard-oriented users and for accessibility purposes.

### Implementation Note
This has been fixed in branch `improvement/add-keyboard-navigation-to-wizard` by adding keyboard event handling to the setup wizard widget.

TheOneRing:
Hi Micheal,
The handling of the Enter key on the url filled was intentionally removed in https://github.com/opencloud-eu/desktop/commit/332964f017af3566981a8bc0ee489f06329445a8 .
It leads to various error cases in the wizard and doesn't follow the default behaviour of the platforms. 

That the buttons don't handle keyboard input is a MacOS limitation, Steve Jobs decided that you must use the mouse.
To enable keyboard navigation, you need to enable it in your accessibility settings.

PR #227, opened 2025-04-22
Improve keyboard navigation in setup wizard

michaelstingl:
## Summary
- Add keyboard navigation support to the setup wizard
- Enter key now triggers Next/Continue/Done buttons
- Escape key now cancels/closes the wizard dialog
- Installs event filters to capture keyboard events from all child widgets
- Fixes issue #226

## Test plan
1. Start the OpenCloud Desktop app
2. Click on \Add Account\
3. Enter a server URL and press Enter - should continue to next step
4. When the login screen appears, press Enter - should open web browser
5. Complete login and verify Enter key works to finish setup
6. Test that Escape key closes the wizard at any stage

### DIFF
--- a/src/gui/newwizard/setupwizardwidget.cpp
+++ b/src/gui/newwizard/setupwizardwidget.cpp
@@ -10,6 +10,7 @@
 #include <QLabel>
 #include <QMessageBox>
 #include <QStyleFactory>
+#include <QKeyEvent>
 
 using namespace std::chrono_literals;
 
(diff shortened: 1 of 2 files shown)


michaelstingl:
I've added another commit that makes the Escape key behavior consistent with the Cancel button.

When a user presses the Escape key, the same confirmation dialog is now shown as when clicking the Cancel button. This is important to warn users that their entered data could be lost if they cancel the wizard.

Issue #245, opened 2025-05-02
Error message in case of insufficient space quota


kulmann:
In case I exceed the default space quota of 1GB I'd appreciate a more precise hint that I was too blind to adjust the space quota. 🤓 

Current message (see screenshot): `There is insufficient space available on the server for some uploads.`

![Image](https://github.com/user-attachments/assets/c4f490cf-e495-4c8a-bbda-aa791b504032)

Suggestion (happy about better ones):

```
Space quota exceeded. Please contact an admin or a manager of this space.
```

PR #248, opened 2025-05-05
Update quota exeeded message

TheOneRing:
Fixes: #245

### DIFF
--- a/src/libsync/syncengine.cpp
+++ b/src/libsync/syncengine.cpp
@@ -831,9 +831,10 @@ void SyncEngine::slotInsufficientLocalStorage()
 
 void SyncEngine::slotInsufficientRemoteStorage()
 {
-    auto msg = tr("There is insufficient space available on the server for some uploads.");
-    if (_uniqueErrors.contains(msg))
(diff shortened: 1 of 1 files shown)


src/libsync/syncengine.cpp
```diff
@@ -831,9 +831,10 @@ void SyncEngine::slotInsufficientLocalStorage()
 
 void SyncEngine::slotInsufficientRemoteStorage()
 {
-    auto msg = tr("There is insufficient space available on the server for some uploads.");
-    if (_uniqueErrors.contains(msg))
+    const QString msg = tr("Space quota exceeded. Please contact the Administrator of this space.");
```
kulmann, line 834: a space has no admin 🙈 only a manager. that's the harder to explain part... space quota can be changed by: a) global admin, b) manager of the space.

TheOneRing: Contact your space manager could also lead to confusion....

kulmann: `Please contact a responsible person` 😆  yeah, let's go with your version. Administrator is easier to understand.

PR #256, opened 2025-05-12
Remove ownCloudGui class forward declaration

prashant-gurung899:
This PR removes `ownCloudGui` class forward declaration.


individual-it:
CC @TheOneRing

TheOneRing:
Hi @prashant-gurung899 thx for your contribution.

The ownCloudGui class, was removed, so the correct fix for this forward declaration would be to remove the declaration altogether.
Please update your pull request and just remove that line.
Thx

PR #263, opened 2025-05-19
Test(refactor): remove vfs related tests

anon-pradip:
This PR removes the vfs related steps and respective methods and variables


TheOneRing:
VFS will come back.

anon-pradip:
> VFS will come back.
Yeah. We can pick these tests from this PR when VFS feature is implemented.

individual-it (approved):
@TheOneRing I would also suggest to merge this and we can bring it back whenever we need it, but it will then not be in our way as long as there is not VFS.
No need to document and remember that its not there

TheOneRing (approved):

PR #266, opened 2025-05-19
Move default exclude file to a resource

TheOneRing:
In an ancient past, this code enabled customers to deploy a custom exclude list.
But on Mac we already use a path from inside the signed bundle.
In AppImage, we were similarly using a file from inside the AppImage.


TheOneRing:
@Narrat https://github.com/opencloud-eu/desktop/pull/268#issuecomment-2899195272

Loading the default exclude file was broken on all platforms but Linux. If you want to modify the excludes for a user, there is an exclude file in the application config dir, which can be modified using the ui.

If you have a scenario where you need to roll out exclude files to multiple users (enterprise setup) please open a feature request.

PR #268, opened 2025-05-19
Require global settings to always be located in /etc/

TheOneRing:


Narrat:
Dunno if the change came with this or #266 
Is it intended, that the dummy `sync-exclude.lst` file isn't installed anymore when building and installing on Linux? If so shouldn't it be documented somewhere where to place such a file if needed? It isn't really clear if it is still `/etc/OpenCloud/` or directly in `/etc/` (which is discouraged).

TheOneRing:
Answered in https://github.com/opencloud-eu/desktop/pull/266#issuecomment-2900226079

PR #276, opened 2025-05-21
add test to login with invalid credentials

prashant-gurung899:
This PR adds test to login a user with invalid credentials.

``` feature
Scenario: login with invalid credentials
```


prashant-gurung899:
Closing this PR as testing login with invalid credentials in not compatible for Opencloud because the login is done via web browser and it gets stuck in the step when invalid credentials are used.

CC @individual-it @anon-pradip

PR #282, opened 2025-05-22
Test(refactor): fix sync path in windows

anon-pradip:
This PR fixes failing scenarios in the the suite `tst_activity` for windows. Test of `Folder1/a\\a.txt` is skipped in windows because we cannot make file with <b>slashes</b> and I could not find the common file pattern for both linux and windows that can be blacklisted.

---

`CLIENT_CONFIG_DIR` was removed because the code no longer needed to dynamically manage or set a separate directory location for the client configuration. We can use fixed paths for configuration files.

---

`settings = QSettings(get_config('clientConfigFile'), QSettings.Format.IniFormat)` &rarr; forces the use of the INI format which in turn makes the config file to have the same structure on all operating systems

---
Moved server cleanup (delete_project_spaces, delete_created_users) to `Order 1` and client cleanup to `Order 2`. This ensures server-side resources are cleaned up while network connections (i.e, socket connections) are still active, preventing cleanup failures due to closed sockets terminating the network connections required for server communication.

---

The accounts tab in the desktop client does not contain the display name for any user. While cleaning up users in windows, the account connection need to be removed, and the selector used the display name which resulted in failure to remove the users. So fixed this problems.

---

Added `'[A-Z]{1}:\/'` pattern to match Windows drive paths using forward slashes (C:/) alongside existing backslash format (C:\\)

---
Used `waitForObject` instead of `waitForObjectExists` because `waitForObject` waits until the `objectOrName` is accessible (i.e., it exists and is visible and enabled); whereas `waitForObjectExists` is for non-visible objects: [waitForObject(name)](https://doc.qt.io/squish/waitforobject-function.html#object-waitforobject-objectorname)

---
<h4> These tests pass on both linux and windows </h4>


test/gui/shared/scripts/pageObjects/SyncConnection.py
```diff
@@ -18,7 +18,6 @@ class SyncConnection:
     FOLDER_SYNC_CONNECTION_MENU_BUTTON = {
         "container": names.quickWidget_scrollView_ScrollView,
         "id": "moreButton",
-        "source": "image://opencloud?theme=fontawesome&icon=&enabled=true&size=undefined",
```
anon-pradip: With this, the locator `FOLDER_SYNC_CONNECTION_MENU_BUTTON` will not be found and removing this does not affect the test so IMHO, we don't need this anymore.

test/gui/shared/steps/account_context.py
```diff
@@ -58,8 +58,10 @@ def step(context, username):
 
     enter_password.login_after_setup(username, password)
 
+    sync_paths = generate_account_config([username])
+
```
prashant-gurung899, line 62:
```suggestion
```

test/gui/tst_activity/test.feature
```diff
@@ -18,20 +18,35 @@ Feature: filter activity for user
...
             | resource         | status      | account                              |
             | Folder1/a\\a.txt | Blacklisted | Alice Hansen@%local_server_hostname% |
+
+    @skipOnLinux
+    Scenario: filter not synced activities (Windows only)
+        Given user "Alice" has been created in the server with default attributes
+        And user "Alice" has set up a client with default settings
+        When user "Alice" creates the following files inside the sync folder:
+            | files             |
+            | /.htaccess        |
```
prashant-gurung899:
```suggestion
            | files      |
            | /.htaccess |
```

test/gui/tst_activity/test.feature
```diff
@@ -18,20 +18,35 @@ Feature: filter activity for user
...
+        And user "Alice" has created a folder "Folder1" inside the sync folder
         When user "Alice" creates the following files inside the sync folder:
             | files             |
             | /.htaccess        |
             | /Folder1/a\\a.txt |
         And the user clicks on the activity tab
         And the user selects "Not Synced" tab in the activity
         Then the file "Folder1/a\\a.txt" should be blacklisted
-        And the file ".htaccess" should be excluded
+        Then the file ".htaccess" should be excluded
```
prashant-gurung899:
```suggestion
        And the file ".htaccess" should be excluded
```

test/gui/shared/scripts/bdd_hooks.py
```diff
@@ -182,10 +181,14 @@ def teardown_client():
...
-        accounts, _ = Toolbar.get_accounts()
-        for account in accounts:
-            Toolbar.open_account(account["displayname"])
+        accounts, selectors = Toolbar.get_accounts()
+        for display_name in selectors:
+            _, account_objects = Toolbar.get_accounts()
+            squish.mouseClick(squish.waitForObject(account_objects[display_name]))
             AccountSetting.remove_account_connection()
+
+        #re-fetch accounts after removing from UI
```
prashant-gurung899: Doesn't matter but...
```suggestion
        # re-fetch accounts after removing from UI
```

individual-it (approved):

PR #294, opened 2025-05-26
docs: Add code comments explaining dual WebFinger calls

### LINKED ISSUE #271: Document why we perform Webfinger twice
For some reason, we perform the Web finger lookup twice.
If that's correct, we should document why we do it.

# Unauthenticated
https://github.com/opencloud-eu/desktop/blob/b3914432e0f2988c0ae77dd8aa0022ad9726b254/src/gui/newwizard/states/serverurlsetupwizardstate.cpp#L65

# Authenticated
https://github.com/opencloud-eu/desktop/blob/b3914432e0f2988c0ae77dd8aa0022ad9726b254/src/gui/newwizard/states/oauthcredentialssetupwizardstate.cpp#L65

michaelstingl:
## Summary
- Added inline documentation to clarify the two WebFinger calls in the authentication flow
- Addresses confusion mentioned in #271

## Details
This PR adds code comments at both WebFinger call locations:
1. **First call** in `serverurlsetupwizardstate.cpp`: Unauthenticated call to discover IdP
2. **Second call** in `oauthcredentialssetupwizardstate.cpp`: Authenticated call to discover user's instance(s)

## Context
As discussed in #271, the dual WebFinger calls are intentional and serve different purposes in OpenCloud's multi-tenant architecture. These comments will help future contributors understand the design.

Fixes #271

### DIFF
--- a/src/gui/newwizard/states/oauthcredentialssetupwizardstate.cpp
+++ b/src/gui/newwizard/states/oauthcredentialssetupwizardstate.cpp
@@ -59,8 +59,12 @@ OAuthCredentialsSetupWizardState::OAuthCredentialsSetupWizardState(SetupWizardCo
             }
         };
 
-        // we run this job here so that it runs during the transition state
-        // sure, it's not the cleanest ever approach, but currently it's good enough
+        // SECOND WEBFINGER CALL (authenticated):
+        // This discovers which OpenCloud instance(s) the authenticated user has access to.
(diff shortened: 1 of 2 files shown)


michaelstingl:
Thanks for the feedback @TheOneRing! 

I've added the link to the backend WebFinger documentation as requested:
- https://github.com/opencloud-eu/opencloud/blob/main/services/webfinger/README.md

The branch has also been rebased on the latest main.

PR #300, opened 2025-05-28
Handle return key for the url wizard page

TheOneRing:


TheOneRing:
@tbsbdr this approach differs in that, if a clearly invalid URL is entered, the next button is disabled and pressing return has no effect. The URL is only validated on a string level, so if you enter a URL that does not point to a valid OpenCloud server, you will still get an error message.

src/gui/newwizard/pages/serverurlsetupwizardpage.cpp
```diff
@@ -4,7 +4,44 @@
...
+{
+    Q_OBJECT
+public:
+    using QValidator::QValidator;
+    State validate(QString &input, int &) const override
+    {
+        if (input.isEmpty()) {
+            return Intermediate;
+        }
+        const QRegularExpression urlRegex(uR"(^(https?:\/\/)?(www\.)?([a-zA-Z0-9-]+\.)+[a-zA-Z]{2,}(\/[^\s]*)?$)"_s);
```
TheOneRing: aaand I found the first issue, this does not allow to enter umlauts

PR #317, opened 2025-06-10
Tests Only: fix add accounts step

anon-pradip:
- [x] Needs https://github.com/opencloud-eu/desktop/pull/282 be merged 

- Refactored path handling throughout for tests.
- **SetupClientHelper.py**
  - Removed forced conversion to forward slashes in `create_user_sync_path` and `create_space_path` returns; now returns native OS paths.
  - Eliminated Windows-specific slash replacement in `get_resource_path` for unified path formatting.
- **SyncHelper.py**
  - In `filter_messages_for_item`, stopped converting backslashes to slashes; now only trims trailing separators.
- **AccountConnectionWizard.py**
  - Appended `syncConnectionName` from config to sync folder path for correct folder (`space name like Personal` in general) selection.
- **account_context.py**
  - On Windows, replaced forward slashes with backslashes in `sync_path` before processing, because `/` is passed from the test step; ensuring platform compatibility

### DIFF
--- a/test/gui/shared/scripts/helpers/SetupClientHelper.py
+++ b/test/gui/shared/scripts/helpers/SetupClientHelper.py
@@ -60,7 +60,7 @@ def create_user_sync_path(username):
         makedirs(user_sync_path)
 
     set_current_user_sync_path(user_sync_path)
-    return user_sync_path.replace('\\', '/')
+    return user_sync_path
 
 
 def create_space_path(username, space='Personal'):
@@ -68,7 +68,7 @@ def create_space_path(username, space='Personal'):
     space_path = join(user_sync_path, space, '')
     if not exists(space_path):
         makedirs(space_path)
-    return space_path.replace('\\', '/')
+    return space_path
 
 
 def set_current_user_sync_path(sync_path):
@@ -83,8 +83,6 @@ def get_resource_path(resource='', user='', space=''):
     sync_path = join(sync_path, space)
     sync_path = join(get_config('clientRootSyncPath'), sync_path)
     resource = resource.replace(sync_path, '').strip('/').strip('\\')
-    if is_windows():
-        resource = resource.replace('/', '\\')
     return join(
         sync_path,
         resource,
--- a/test/gui/tst_addAccount/test.feature
+++ b/test/gui/tst_addAccount/test.feature
@@ -45,6 +45,10 @@ Feature: adding accounts
         When the user adds the server "%local_server%"
         And the user accepts the certificate
         Then credentials wizard should be visible
+        When the user adds the following account:
+            | user     | Alice |
+            | password | 1234  |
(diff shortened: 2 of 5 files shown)


test/gui/shared/scripts/helpers/WinPipeHelper.py
```diff
@@ -59,8 +59,6 @@ def connect_to_pipe_server(self):
 
             self.connected = True
 
-            self.sendCommand('VERSION:\n')
-            self.sendCommand('GET_STRINGS:\n')
```
saw-jan: this is not really required. :+1:

test/gui/shared/scripts/helpers/SetupClientHelper.py
```diff
@@ -60,15 +60,15 @@ def create_user_sync_path(username):
         makedirs(user_sync_path)
 
     set_current_user_sync_path(user_sync_path)
-    return user_sync_path.replace('\\', '/')
+    return user_sync_path
```
anon-pradip, line 63: This is done to use the path given by the system and not trying to manually change the path unless required.

test/gui/tst_addAccount/test.feature
```diff
@@ -45,6 +45,10 @@ Feature: adding accounts
         When the user adds the server "%local_server%"
         And the user accepts the certificate
         Then credentials wizard should be visible
+        When the user adds the following account:
+            | user     | Alice   |
+            | password | 1234    |
+        Then "Alice Hansen" account should be opened
```
anon-pradip: To close the desktop client app, we should be navigated to the screen showing user accounts then we can remove the connections. This is not the optimal way, we should manage in the hooks.

test/gui/tst_addAccount/test.feature
```diff
@@ -45,6 +45,10 @@ Feature: adding accounts
         When the user adds the server "%local_server%"
         And the user accepts the certificate
         Then credentials wizard should be visible
+        When the user adds the following account:
+            | user     | Alice   |
+            | password | 1234    |
```
prashant-gurung899:
```suggestion
            | user     | Alice |
            | password | 1234  |
```

test/gui/shared/scripts/helpers/WinPipeHelper.py
individual-it: please pull this change into a separate PR

individual-it (changes requested):

test/gui/shared/scripts/helpers/WinPipeHelper.py
anon-pradip: Made a separate PR: https://github.com/opencloud-eu/desktop/pull/324

individual-it (approved):

PR #328, opened 2025-06-19
Test(refactor): Clean user in windows

anon-pradip:
## Description
This PR enhances the test teardown process to properly clean up user accounts in Windows OS, particularly for scenarios where the test doesn't reach the Account Settings page.

### Concerned issue: ###
In test scenarios where the test flow doesn't navigate to the Account Settings page (e.g., during advanced configuration checks), the test teardown fails to clean up user accounts properly. This occurs because the cleanup logic expects to be in the Account Settings context.

### What's done to solve the issue: ###
- The sole purpose is to remove account connection in those type of screens.
- For those kind of screens, if there is still an active widget, we get that from `get_active_widget()`. And finally if that active widget is not a `SetupWizardWidget`, then we remove account connection through UI.

### DIFF
--- a/test/gui/shared/scripts/bdd_hooks.py
+++ b/test/gui/shared/scripts/bdd_hooks.py
@@ -18,6 +18,7 @@
 import shutil
 import os
 from datetime import datetime
+from types import SimpleNamespace
 
 from helpers.StacktraceHelper import get_core_dumps, generate_stacktrace
 from helpers.SyncHelper import close_socket_connection, clear_waited_after_sync
@@ -38,6 +39,7 @@ from helpers.ReportHelper import save_video_recording, take_screenshot, is_video
 from pageObjects.Toolbar import Toolbar
 from pageObjects.AccountSetting import AccountSetting
 from pageObjects.AccountConnectionWizard import AccountConnectionWizard
+import names
 
 # Squish test settings:
 # This controls whether a test (scenario) should stop execution on failure or not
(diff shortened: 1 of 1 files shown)


test/gui/shared/scripts/bdd_hooks.py
```diff
@@ -172,6 +173,16 @@ def hook(context):
     PREVIOUS_ERROR_RESULT_COUNT = test.resultCount("errors")
 
 
+def get_active_widget():
+    ch = object.children(squish.waitForObject(AccountSetting.DIALOG_STACK, 500))
```
individual-it: better to use pre-defined timeout from https://github.com/opencloud-eu/desktop/blob/d70659393da0f2d90f193e94c2ca2078d615bdd9/test/gui/shared/scripts/helpers/ConfigHelper.py#L68-L70 than having a magic number here

test/gui/shared/scripts/bdd_hooks.py
```diff
@@ -172,6 +173,16 @@ def hook(context):
...
+    for obj in ch:
```
individual-it: `obj` and `ch` are not good descriptive names

test/gui/shared/scripts/bdd_hooks.py
```diff
@@ -181,16 +192,18 @@ def teardown_client():
...
-            _, account_objects = Toolbar.get_accounts()
-            squish.mouseClick(squish.waitForObject(account_objects[display_name]))
-            AccountSetting.remove_account_connection()
-
-        # re-fetch accounts after removing from UI
-        accounts, _ = Toolbar.get_accounts()
-        if accounts:
-            squish.waitForObject(AccountConnectionWizard.SERVER_ADDRESS_BOX)
+        active_widget = get_active_widget()
+        if active_widget.objectName and active_widget.objectName != "SetupWizardWidget":
```
individual-it: can we take that name from https://github.com/opencloud-eu/desktop/blob/d70659393da0f2d90f193e94c2ca2078d615bdd9/test/gui/shared/scripts/names.py#L32 ?
Its better to have all the names in one place

anon-pradip:
As these changes are not required and covered in: https://github.com/opencloud-eu/desktop/pull/331, closing this PR.

anon-pradip:
These changes are quick fix for now. We will need more better and enhanced solution that works with upcoming future changes as well.

individual-it (approved):

PR #333, opened 2025-06-24
Enable http2 support by default

TheOneRing:
Fixes: #27

### DIFF
--- a/src/libsync/owncloudpropagator.cpp
+++ b/src/libsync/owncloudpropagator.cpp
@@ -76,7 +76,7 @@ int OwncloudPropagator::maximumActiveTransferJob()
         // disable parallelism when there is a network limit.
         return 1;
     }
-    return qMin(3, qCeil(_syncOptions._parallelNetworkJobs / 2.));
+    return std::max(3, qCeil(_syncOptions._parallelNetworkJobs() / 2.));
 }
 
 /* The maximum number of active jobs in parallel  */


src/libsync/owncloudpropagator.cpp
```diff
@@ -76,15 +76,15 @@ int OwncloudPropagator::maximumActiveTransferJob()
         // disable parallelism when there is a network limit.
         return 1;
     }
-    return qMin(3, qCeil(_syncOptions._parallelNetworkJobs / 2.));
+    return std::max(3, qCeil(_syncOptions._parallelNetworkJobs() / 2.));
```
TheOneRing, line 79: @dragotin the min must have been an error because it would basically always set it to 3... besides the unit tests where _parallelNetworkJobs is set to 1 🙈

src/libsync/account.h
```diff
@@ -238,7 +238,7 @@ public Q_SLOTS:
     QPointer<QNetworkDiskCache> _networkCache = nullptr;
     QPointer<ResourcesCache> _resourcesCache;
     QScopedPointer<AbstractCredentials> _credentials;
-    bool _http2Supported = false;
+    bool _http2Supported = true;
```
guruz, line 241: Is the goal not to initialize with false and then have https://github.com/opencloud-eu/desktop/pull/333/files#diff-12d910d5c97c3d9989a6731c12f2a81fb9a1153063bfe76f560afcc82b6420b4L67 set it to true?

supported = we tried it and it was actually used?
allowed = we attempt to try it at all

So basically only set 20 parallel if it worked once as http2

guruz: Note: I don't know how often  Folder::loadSyncOptions() gets called, because then the opt._parallelNetworkJobs  would need to be re-evaluated after we know if works (if it gets called only once)

TheOneRing: that's why I replaced it with a std::function, the alternative would be to update it once _http2Supported changed.
The value here shouldn't really matter, as we will definitely have checked for http2 support once a sync starts.

src/libsync/syncoptions.h
```diff
@@ -43,7 +44,7 @@ class OPENCLOUD_SYNC_EXPORT SyncOptions
     QSharedPointer<Vfs> _vfs;
 
     /** The maximum number of active jobs in parallel  */
-    int _parallelNetworkJobs = 6;
+    std::function<int()> _parallelNetworkJobs = [] { return 6; };
```
guruz, line 47: hm, to make it more easy to understand, i'd add a comment here: // will be overwriten later in loadSyncOptions()

Issue #344, opened 2025-06-26
Migration from owncloud client to opencloud client fails


jnweiger:
Ubuntu 24.10.
opencloud 2.0.0-rc4 appimage

I have an old ~/.config/ownCloud/owncloud.cfg from ownCloud ownCloud 5.2.1
with two servers configured.

The new opencloud desktop client does not pick up this config, it creates a new empty ~/.config/OpenCloud/opencloud.cfg
Expected behaviour: as opencloud is the successor of owncloud, it should detect the old config and offer a migration step.

A manual `cp ~/.config/ownCloud/owncloud.cfg ~/.config/OpenCloud/opencloud.cfg` does not work:

    Einige Einstellungen wurden in einer neueren Version dieses Clienten konfiguriert und nutzen Funktionen, die in dieser Version nicht verfügbar sind.

Side note: This message should not say older version instead of newer version.

TheOneRing:
Not supported

PR #345, opened 2025-06-26
showDowngradeDialog() appears not only on downgrades

jnweiger:
grafik

When the .cfg file was copied over from a nextcloud or owncloud client, this message may also happen. (To be expected).

The current language implies that we are an older version now, which is misleading.  Suggest to use something more neutral, e.g.  'different' instead.

See also: #344


TheOneRing:
Its not expected that someone puts a random file in the config dir and expects it to work.
The scenario is just not supported and in this case, lowers the value of the message for actual opencloud users.

PR #347, opened 2025-06-26
Update desktop_de.ts Fesnter -> Fenster

jnweiger:
Fix a typo

grafik


TheOneRing:
Also fixed on transifxe but that will only end up on main.

Issue #354, opened 2025-06-30
Incorrect user page. error message doesn't contain user name


ScharfViktor:
version https://github.com/opencloud-eu/desktop/releases/tag/v2.0.0-rc.4

Steps: 
- start opencloud
- `alan` connects their account to the desktop and creates data
- `admin` deletes `alan` and creates a new `alan` with the same credentials
- new `alan` connects to the desktop using the old account

or `docker compose down -v` and start again `docker compose up -v`

Actual: 
![Image](https://github.com/user-attachments/assets/5267190e-ea79-4c95-b893-c4cc4544da22)

in this case restart doesn't work and error message doesn't help to understand problem

TheOneRing:
Hmm I can fix the error message, but in this specific case you wold get.
"You logged in with user Alan Turin, but mist log in with Alan Turing" because the use name is the same but the id changed.

ScharfViktor:
> Hmm I can fix the error message, but in this specific case you wold get. "You logged in with user Alan Turin, but mist log in with Alan Turing" because the use name is the same but the id changed.

I believe this message will be relevant and useful if the user has two accounts: “You are logged in as Alan Turin, but you should log in as Dennis Ritchie Turing.”

rhafer:
> > Hmm I can fix the error message, but in this specific case you wold get. "You logged in with user Alan Turin, but mist log in with Alan Turing" because the use name is the same but the id changed.

That isn't exactly a helpful error message either, I guess. But it's a rather specific case. A user being re-created with the same username but a different id that is then trying to login using the same device as the old sounds highly unlikely. (Or just like  some kind of development/test setup)

> I believe this message will be relevant and useful if the user has two accounts: “You are logged in as Alan Turin,but mist log in with Dennis Ritchie Turing.”

PR #355, opened 2025-06-30
OAuth: Only display user name in an error if we have one

TheOneRing:
The OpenCloud build in idp does not provide a prefered_username
Fixes: #354

### DIFF
--- a/src/libsync/creds/oauth.cpp
+++ b/src/libsync/creds/oauth.cpp
@@ -374,12 +374,21 @@ void OAuth::startAuthentication()
                         reportError(tr("The audience of the id_token did not contain \"%1\"").arg(_clientId));
                     } else if (_idToken.isValid() && _idToken.sub() != idToken.sub()) {
                         // Connected with the wrong user
-                        qCWarning(lcOauth) << "We expected the user" << _idToken.preferred_username() << "but the server answered with user"
-                                           << idToken.preferred_username();
-                        const QString message = tr("<h1>Incorrect user</h1>"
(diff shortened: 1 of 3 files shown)


TheOneRing:
@ScharfViktor is the new message that will be displayed better?

src/libsync/creds/oauth.cpp
```diff
@@ -374,12 +374,21 @@ void OAuth::startAuthentication()
...
-                                           << idToken.preferred_username();
-                        const QString message = tr("<h1>Incorrect user</h1>"
-                                                   "<p>You logged-in with user <em>%1</em>, but must login with user <em>%2</em>.<br>"
-                                                   "Please return to the %3 and restart the authentication.</p>")
-                                                    .arg(idToken.preferred_username(), _idToken.preferred_username(), Theme::instance()->appNameGUI());
+                        qCWarning(lcOauth) << "We expected the user" << _idToken.toJson() << "but the server answered with user" << idToken.toJson();
+                        const QString expectedName = !_idToken.preferred_username().isEmpty() ? _idToken.preferred_username() : _idToken.name();
+                        const QString actualName = !idToken.preferred_username().isEmpty() ? idToken.preferred_username() : idToken.name();
+                        QString message;
+                        if (!expectedName.isEmpty() && !actualName.isEmpty()) {
```
rhafer: To avoid message like "You logged in with user Alan Turin, but mist log in with Alan Turing" (like mentioned in https://github.com/opencloud-eu/desktop/issues/354), would it make sense to add a `&&  (expectedName != actualName)` condition here?

TheOneRing: absolute corner case, but a good idea

PR #360, opened 2025-07-02
Test(refactor): Edit manual test plan file

anon-pradip:
This PR refactors the manual test plan file. Removed some steps because they are not relevant (outdated) for current desktop client status.
Added a check: `Viewer creates folder in a project space` for a folder to be blacklisted in `8. Spaces and Sharing`


anon-pradip:
> Also add test-suite for `Manager can add new folders in Space.` We have it automated in `tst_spaces`
Done both :rocket:

Issue #370, opened 2025-07-09
[QA] Automate additional test cases in the manual test plan


anon-pradip:
Several test cases in our manual test plan: https://github.com/opencloud-eu/desktop/blob/main/test/manual/testplan.md could be automated to improve efficiency and test coverage in our CI pipeline. The following test cases are good candidates for automation:

---

### **1. Install, Setup and Configuration**

* [x] Configure sync folder https://github.com/opencloud-eu/desktop/pull/539
    https://github.com/opencloud-eu/desktop/blob/dc4d5e07697352dd7daa8b448e21440c9e262019/test/gui/tst_addAccount/test.feature#L54
* [x] Re-add an account @anon-pradip https://github.com/opencloud-eu/desktop/issues/435  https://github.com/opencloud-eu/desktop/pull/471

### **3. Files** @anon-pradip 

~* [ ] Copy multiple files at a time~ => Since selecting multiple files at once is not possible while testing
* [x] Create and delete a file with special characters in the name https://github.com/opencloud-eu/desktop/pull/528
* [x] Upload a 50MB file https://github.com/opencloud-eu/desktop/pull/530
* [x] Upload multiple files (with a sum of 1000MB) https://github.com/opencloud-eu/desktop/pull/536

### **4. Move Files and Folders**  

* [x] Move from sub-levels to root https://github.com/opencloud-eu/desktop/pull/537
* [x] Move files from one folder to another @anon-pradip https://github.com/opencloud-eu/desktop/pull/390

### **5. Edit Files**

* [x] Replace a `.pdf` file @anon-pradip  https://github.com/opencloud-eu/desktop/pull/394

### **6. Delete Files and Folders**

* [x] Delete multiple files @anon-pradip https://github.com/opencloud-eu/desktop/pull/424
~* [ ] Delete a large file (2048MB)~ => Since check for 1GB file is already done, check for 2GB file is better to be tested manually   => Reason: Not making tests and CI be more overloaded and more time consuming

### **8. Spaces and Sharing**

* [x] Editor can rename files in Space https://github.com/opencloud-eu/desktop/pull/481
* [x] Remove a project space sync connection https://github.com/opencloud-eu/desktop/pull/491
* [ ] Sync a received shared folder (can view - secure) 
         &rarr; Files can't be downloaded by the client due to `Secure viewer` permission role so automation might not be possible (holding this)
* [x] Sync a received shared folder (can view) https://github.com/opencloud-eu/desktop/pull/506
[...]

PR #380, opened 2025-07-10
Test: add test for re-adding an account

anon-pradip:
Added a new scenario for re-adding an account after removing it:
```gherkin
Scenario: Re-add an account
```
Part of: https://github.com/opencloud-eu/desktop/issues/370


test/gui/tst_addAccount/test.feature
```diff
@@ -75,3 +75,18 @@ Feature: adding accounts
...
         Then the button to open sync connection wizard should be disabled
+
+
+    Scenario: Re-add an account
+        Given user "Alice" has created folder "large-folder" in the server
+        And user "Alice" has uploaded file with content "test content" to "testFile.txt" in the server
+        And user "Alice" has set up a client with default settings
+        And the user has removed the connection for user "Alice" and host %local_server_hostname%
+        When the user opens the add-account dialog
+        When the user adds the following account:
```
prashant-gurung899:
```suggestion
        When the user opens the add-account dialog
        And the user adds the following account:
```

test/gui/shared/steps/account_context.py
```diff
@@ -284,3 +284,12 @@ def step(context, warn_message):
...
     )
+
+
+@Given('the user has removed the connection for user "|any|" and host |any|')
+def step(context, username, _):
+    displayname = get_displayname_for_user(username)
+    displayname = substitute_inline_codes(displayname)
+
+    Toolbar.open_account(displayname)
+    AccountSetting.remove_account_connection()
```
individual-it: this is duplicated code

test/gui/shared/steps/account_context.py
```diff
@@ -284,3 +284,12 @@ def step(context, warn_message):
         warn_message in actual_message,
         'Contains warning message',
     )
+
+
+@Given('the user has removed the connection for user "|any|" and host |any|')
+def step(context, username, _):
```
individual-it: if the hostname is not needed, why do we even have it in the feature file?

anon-pradip: For this `Given` step, I removed the unused `host` parameter here in this PR. But for other steps and step definitions please refer: https://github.com/opencloud-eu/desktop/pull/413. And lets wait for #413 be merged first

PR #387, opened 2025-07-11
enhancement(ci): sync translations

fschade:
## Description
make use of woodpecker crons to sync the transifex translations, this needs https://github.com/opencloud-eu/woodpecker-ci-config-service/releases v1.0.1 or newer.

## Related Issue
- https://github.com/opencloud-eu/qa/issues/48


.woodpecker/translation.yaml
```diff
@@ -0,0 +1,44 @@
...
+      - apt update
+      - ln -fs /usr/share/zoneinfo/Europe/Berlin /etc/localtime
+      - apt install -y qt6-l10n-tools make git curl
+      - rm translations/*.ts
+      - git checkout translations/desktop_en.ts
+      - /usr/lib/qt6/bin/lupdate src -no-obsolete -ts translations/desktop_en.ts
+      - curl -o- https://raw.githubusercontent.com/transifex/cli/master/install.sh | bash
+      - . ~/.profile
+      - tx push -s
+      - rm -rf /tmp/tx
```
TheOneRing, line 23: Please add the removal step from the old script.
If a translation falls below the 75% threshold, it should be removed from the code base, in the past we carried outdated unmaintained translations with us, that where about 11 years outdated.

TheOneRing: sorry I apparently can't read.

fschade: 🤣 ... the friday syndrome

PR #394, opened 2025-07-15
Test: add test for replacing a pdf file

anon-pradip:
This PR adds test for replacing a `.pdf` file, changing its content multiple times and checking the final content in the sever:
```gherkin
Scenario: Replace and modify the content of a pdf file multiple times
```
- Modified the `get_file_content` function:

    Changed return value based on file type:  `response.content` is to handle binary files (like PDF) correctly.

- Added helper function `copy_resource` to abstract file/folder copying logic.
- Added new step definitions to cover copying files/folders in the sync folder, relevant for the new PDF replacement scenario.

- Improved content verification logic:

    - For `.txt` files, still compares text content.
    - For other files (e.g., PDF), now compares server content byte-for-byte with the local file, ensuring accurate binary file testing.

- test/gui/tst_editFiles/test.feature:
  - Copies an initial PDF to the sync folder.
  - Sequentially replaces it with three different PDFs, waiting for sync after each replacement.
  - Finally checking the content of local file with that of the file in sever

Part of: https://github.com/opencloud-eu/desktop/issues/370


test/gui/tst_editFiles/test.feature
```diff
@@ -24,4 +24,24 @@ Feature: edit files
...
\ No newline at end of file
+        Then as "Alice" the file "testfile.txt" should have the content "overwrite openCloud test text file" in the server
+
+
+   Scenario: Replace and modify the content of a pdf file multiple times
+     Given user "Alice" has set up a client with default settings
+     And user "Alice" has created a file "file1.pdf" with the following content inside the sync folder
+            """
+            to be replaced
+            """
```
individual-it: This will not result in a PDF file, but in a simple ASCII text file with `.pdf` in it's file name.
What do we want to test?

test/gui/shared/steps/file_context.py
```diff
@@ -373,3 +373,25 @@ def step(context, folder_name):
...
     remember_path(f'{folder_path} (2)')
+
+
+@Given(
+    'user "|any|" has created a file "|any|" with the following content inside the sync folder'
+)
+def step(context, username, filename):
+    file_content = '\n'.join(context.multiLineText)
+    file = get_resource_path(filename, username)
+    wait_and_write_file(convert_path_separators_for_os(file), file_content)
```
individual-it: this is a 100% copy of https://github.com/opencloud-eu/desktop/blob/4035b6cf22f6d092ff391813b1c40c32da6c98c3/test/gui/shared/steps/file_context.py#L106-L109
Please do no repeat existing code

test/gui/tst_editFiles/test.feature
```diff
@@ -24,4 +24,19 @@ Feature: edit files
...
\ No newline at end of file
+        Then as "Alice" the file "testfile.txt" should have the content "overwrite openCloud test text file" in the server
+
+
+   Scenario: Replace and modify the content of a pdf file multiple times
+     Given user "Alice" has uploaded file "simple.pdf" to "/simple.pdf" in the server
+     And user "Alice" has created folder "simple-folder" in the server
+     And user "Alice" has uploaded file "simple.pdf" to "simple-folder/simple.pdf" in the server
+     And user "Alice" has set up a client with default settings
+     And the user has overwritten the file "simple-folder/simple.pdf" with content "overwrite openCloud test pdf file"
```
individual-it: why is this step needed?

anon-pradip: Because in upcoming step we need to replace it with different content so changing the content here

individual-it: why? the file has already "content", you have uploaded it in line 33

anon-pradip: Yeah but in line 33, both the files `simple-folder/simple.pdf` and `/simple.pdf` have same content.  But according to 2nd point in the plan:
```
1. Create a .pdf file in the local sync folder
2. Replace it with a same name pdf file having different content
3. Wait for it to sync
4. Modify the .pdf file and add more content
5. Wait for it to sync
6. Modify the .pdf file and add more content
7. Wait for it to sync
```
 `simple-folder/simple.pdf` should have different content so that we can replace `/simple.pdf` file with different content (content of `simple-folder/simple.pdf` after overwriting) but with same file name, which is done in line 36

injecting different content &rarr; line 35
replacing with that different content &rarr; line 36

individual-it: the test says that the new version should also be a PDF file, you are still replacing PDF files with ASCII-text files

test/gui/shared/steps/file_context.py
```diff
@@ -373,3 +376,19 @@ def step(context, folder_name):
     remember_path(folder_path)
     # when account is added, folder with suffix will be created
     remember_path(f'{folder_path} (2)')
+
+
+@Given(
+    r'the user has copied (file|folder) "([^"]*)" to "([^"]*)" in the sync folder',
```
individual-it:
```suggestion
    r'the user has copied (file|folder) "([^"]*)" from outside the sync folder to "([^"]*)" in the sync folder',
```
similar in the When step

test/gui/shared/steps/file_context.py
```diff
@@ -100,6 +101,20 @@ def add_copy_suffix(resource_path, resource_type):
     return resource_path + ' - Copy'
 
 
+def copy_resource(resource_type, source, destination, is_upload_file=False):
+    wait_for_client_to_be_ready()
+    if resource_type == 'file' and is_upload_file:
```
individual-it: can a folder be copied from the "files-for-upload" into the sync folder?

test/gui/shared/steps/file_context.py
```diff
@@ -100,6 +101,20 @@ def add_copy_suffix(resource_path, resource_type):
     return resource_path + ' - Copy'
 
 
+def copy_resource(resource_type, source, destination, is_upload_file=False):
```
individual-it:
```suggestion
def copy_resource(resource_type, source, destination, from_files_for_upload=False):
```

test/gui/shared/steps/server_context.py
```diff
@@ -28,12 +29,26 @@ def step(context, user_name, resource_name):
...
+        # for other types of file, compare local content with server content
+        local_path = get_resource_path(file_name)
+        with open(local_path, "rb") as f:
+            local_bytes = f.read()
+
+        test.compare(
+            bytes_from_server,
+            local_bytes,
+            f"File '{file_name}' should match local copy byte-for-byte"
+        )
```
individual-it: I don't get how that can work.
1. the `file_name` here is the file_name on the server, there might or might not be a file with the same name in the `files-for-upload` folder
2. for comparing with a local file the step definition is super confusing.

individual-it (changes requested):

PR #391, opened 2025-07-14
Brin back VFS related tests

individual-it:
This reverts commit b3914432e0f2988c0ae77dd8aa0022ad9726b254.

PR #395, opened 2025-07-16
bring back some VFS tests

individual-it:
part of #391 just making those tests run that should already work


test/gui/shared/scripts/helpers/SetupClientHelper.py
```diff
@@ -169,7 +169,10 @@ def generate_account_config(users, space='Personal'):
         settings.setValue("localPath", sync_path)
         settings.setValue("paused", 'false')
         settings.setValue("priority", '50')
-        settings.setValue("virtualFilesMode", 'off')
+        if is_windows():
+            settings.setValue("virtualFilesMode", 'cfapi')
+        else:
+            settings.setValue("virtualFilesMode", 'off')
```
prashant-gurung899, line 183:
```suggestion
        settings.setValue("virtualFilesMode", 'off')
        if is_windows():
            settings.setValue("virtualFilesMode", 'cfapi')
```

PR #420, opened 2025-07-28
Remove oc10 data-fingerprint support 

TheOneRing:


butonic:
related https://github.com/opencloud-eu/opencloud/issues/1284

PR #451, opened 2025-08-04
fix: make translation source strings better readable

kulmann:
Tried to create easier readable sentences from some translation source strings.

I discovered an inconsistency: file/folder names and paths were sometimes quoted with single quotes, sometimes not. Usually I think it's easier readable to quote file/folder names and paths. But the majority of string didn't have them quoted, so I removed the single quotes... ok to add the single quotes everywhere? In web we're using double arrows `»file name«` - should we use the same for desktop?

### DIFF
--- a/src/gui/folder.cpp
+++ b/src/gui/folder.cpp
@@ -153,7 +153,7 @@ Result<void, QString> Folder::checkPathLength(const QString &path)
 #ifdef Q_OS_WIN
     if (path.size() > MAX_PATH) {
         if (!FileSystem::longPathsEnabledOnWindows()) {
-            return tr("The path %1 is too long. Please enable long paths in the Windows settings or choose a different folder.").arg(path);
+            return tr("The path »%1« is too long. Please enable long paths in the Windows settings or choose a different folder.").arg(path);
         }
     }
 #else
@@ -195,19 +195,19 @@ bool Folder::checkLocalPath()
         if (error.isEmpty()) {
             qCDebug(lcFolder) << u"Checked local path ok";
             if (!_journal.open()) {
-                error = tr("Failed to open the database for %1.").arg(_definition.localPath());
+                error = tr("Failed to open the database for »%1«.").arg(_definition.localPath());
             }
         }
     } else {
         // Check directory again
         if (!FileSystem::fileExists(_definition.localPath(), fi)) {
-            error = tr("Local folder %1 does not exist.").arg(_definition.localPath());
+            error = tr("Local folder »%1« does not exist.").arg(_definition.localPath());
         } else if (!fi.isDir()) {
-            error = tr("%1 should be a folder but is not.").arg(_definition.localPath());
+            error = tr("»%1« should be a folder but is not.").arg(_definition.localPath());
         } else if (!fi.isReadable()) {
-            error = tr("%1 is not readable.").arg(_definition.localPath());
+            error = tr("»%1« is not readable.").arg(_definition.localPath());
         } else if (!fi.isWritable()) {
-            error = tr("%1 is not writable.").arg(_definition.localPath());
+            error = tr("»%1« is not writable.").arg(_definition.localPath());
         }
     }
     if (!error.isEmpty()) {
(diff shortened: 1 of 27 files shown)


src/gui/updater/ocupdater.cpp
```diff
@@ -377,7 +377,7 @@ void WindowsUpdater::versionInfoArrived(const UpdateInfo &info)
 void WindowsUpdater::showNewVersionAvailableWidget(const UpdateInfo &info)
 {
     // if the version tag is set, there is a newer version.
-    QString txt = tr("<p>A new version of the %1 Client is available.</p>"
+    QString txt = tr("<p>A new version of the %1 Desktop is available.</p>"
```
TheOneRing: Desktop App?
Desktop on its own sounds a bit lost in a sentence.

src/gui/folder.cpp
```diff
@@ -153,7 +153,7 @@ Result<void, QString> Folder::checkPathLength(const QString &path)
 #ifdef Q_OS_WIN
     if (path.size() > MAX_PATH) {
         if (!FileSystem::longPathsEnabledOnWindows()) {
-            return tr("The path '%1' is too long. Please enable long paths in the Windows settings or choose a different folder.").arg(path);
+            return tr("The path %1 is too long. Please enable long paths in the Windows settings or choose a different folder.").arg(path);
```
TheOneRing: A path might contain spaces so in general quoting might make sense?

src/libsync/propagatedownload.cpp
```diff
@@ -456,7 +456,7 @@ void PropagateDownloadFile::start()
         // do a klaas' case clash check.
         if (auto clash = propagator()->localFileNameClash(_item->localName())) {
             done(SyncFileItem::NormalError,
-                tr("File %1 can not be downloaded because of a local file name clash with %2!")
+                tr("The file%1 can not be downloaded because of a local file name clash with %2!")
```
TheOneRing: missing space

src/libsync/propagatedownload.cpp
```diff
@@ -539,7 +539,7 @@ void PropagateDownloadFile::startDownload()
     // do a klaas' case clash check.
     if (auto clash = propagator()->localFileNameClash(_item->localName())) {
         done(SyncFileItem::NormalError,
-            tr("File %1 can not be downloaded because of a local file name clash with %2!")
+            tr("The file%1 can not be downloaded because of a local file name clash with %2!")
```
TheOneRing: same

src/libsync/propagatorjobs.cpp
```diff
@@ -231,7 +231,7 @@ void PropagateLocalRename::start()
             // Fixme: the file that is the reason for the clash could be named here,
             // it would have to come out the localFileNameClash function
             done(SyncFileItem::NormalError,
-                tr("File %1 can not be renamed to %2 because of a local file name clash")
+                tr("The file%1 can not be renamed to %2 because of a local file name clash")
```
TheOneRing: space

TheOneRing:
In one of the unit tests, we check for one of those strings.
```
FAIL!  : TestRemoteDiscovery::testRemoteDiscoveryError(Vfs::Off:400) Compared values are not the same
   Actual   (errorSpy[0][0].toString())                      : "Server replied with an error while reading directory B : Internal Server Fake Error"
   Expected (QString(fatalErrorPrefix + expectedErrorString)): "Server replied with an error while reading directory 'B' : Internal Server Fake Error"
   Loc: [/Users/runner/work/desktop/desktop/test/testremotediscovery.cpp(143)]

TheOneRing:
> @TheOneRing what's your opinion about quoting ALL paths and file/folder names? Wrote something in the PR description, don't know if you saw that.

Didn't see that. Yes, I guess that would look good, but it would increase the complexity of the pr.

kulmann:
> > @TheOneRing what's your opinion about quoting ALL paths and file/folder names? Wrote something in the PR description, don't know if you saw that.
> 
> Didn't see that. Yes, I guess that would look good, but it would increase the complexity of the pr.

Updated the PR. Please check if you like it. Otherwise it's now easy to change with search+replace, so please be honest. 😅

TheOneRing:
In some strings, you removed the <i> in others you kept it. Was that intentional?

kulmann:
> In some strings, you removed the in others you kept it. Was that intentional?

No, that was not intentional, thanks for spotting that. I removed all `<i>` tags now, as using them + quotes doesn't seem necessary.

PR #474, opened 2025-08-08
provide more and better log output to developers after CI runs

individual-it:
1. When we hit an abort in out code, we write the last lines of logs to  $TMPDIR/OpenCloud-crash.log it could be beneficial to dump those lines to the ci logs when the test failed and the log exists.
2. write a separate log folder per feature file and log file per scenario
3. upload normal logs to S3, so that they can be easier accessed

fixes https://github.com/opencloud-eu/desktop/issues/473


individual-it:
@TheOneRing can you try to provoke a crash in this PR to see if it actually works?

TheOneRing:
yay, now it could make sense if I'd keep the log files around and we cat all `/woodpecker/desktop/test/gui/tmp/OpenCloud-crash*` files?

PR #489, opened 2025-08-12
Implement a beta channel branding

TheOneRing:
As we don't support downgrades, switching from a 3.00 pre-release back to a stable release would not be possible.
Therefore, we decided to make the preview clients use an independent configuration, they can also run in parallel with a production client.
The beta client also has asserts enabled so it might crash occasionally, if it does, please open a bug report, the reason for the crash will be in the log file and in `"$TMPDIR/OpenCloud Beta-crash.log"`.


flimmy:
lgtm, please fix linux ci before merge :)

PR #506, opened 2025-08-19
test: add test for syncing a received folder

anon-pradip:


test/gui/shared/scripts/helpers/api/webdav_helper.py
```diff
@@ -79,3 +95,57 @@ def delete_resource(user, resource):
     url = get_resource_path(user, resource)
     response = request.delete(url, user=user)
     assert response.status_code == 204, f"Could not delete folder '{resource}'"
+
+
+def get_resource_id(user, resource):
```
individual-it, line 100: what about using xpath to find the element? https://docs.python.org/3/library/xml.etree.elementtree.html#example

test/gui/shared/scripts/pageObjects/SyncConnection.py
```diff
@@ -142,3 +151,16 @@ def confirm_folder_sync_connection_removal():
...
+    def get_permission_error_message():
+        # Wait for permission error label to appear
+        result = squish.waitFor(
+            lambda: object.exists(SyncConnection.PERMISSION_ERROR_LABEL),
+            SyncConnection.PERMISSION_ERROR_WAIT_TIMEOUT * 1000
+        )
+        if result:
+            return str(squish.waitForObject(SyncConnection.PERMISSION_ERROR_LABEL, get_config("lowestSyncTimeout") * 1000).text)
+        else:
+            return ""
```
individual-it: in that case there seems to be an error, so throw an exception

anon-pradip: But in https://github.com/opencloud-eu/desktop/blob/dd6e432652731943dc3d5b38fb34f934f027ffe2/test/gui/shared/steps/sync_context.py#L309 we need to assert with empty string, so throwing an error in line 166 will make test fail

test/gui/shared/scripts/helpers/api/webdav_helper.py
```diff
@@ -35,12 +42,21 @@ def get_file_content(user, resource):
     return response.content
 
 
+def get_propfind_root_element(user, resource):
```
saw-jan:
```suggestion
def get_folder_items(user, resource):
```

test/gui/shared/scripts/helpers/api/webdav_helper.py
```diff
@@ -79,3 +95,53 @@ def delete_resource(user, resource):
...
+
+
+def get_resource_id(user, resource):
+    root_element = get_propfind_root_element(user, resource)
+
+    # Finding the fileid elements using XPath with namespace notation
+    fileid_elements = root_element.findall(".//{http://owncloud.org/ns}fileid")
+
+    # Extracting the text from the first fileid element
+    if fileid_elements:
```
saw-jan:
```suggestion
    # First element is the desired resource's file id
    if fileid_elements:
```

test/gui/shared/scripts/pageObjects/SyncConnectionWizard.py
```diff
@@ -115,6 +115,7 @@ class SyncConnectionWizard:
         "visible": 1,
     }
 
+
```
saw-jan:
```suggestion
```

test/gui/shared/steps/sync_context.py
```diff
@@ -296,3 +296,23 @@ def step(context, should_or_should_not):
...
+def step(context):
+    expected_error_message = '\n'.join(context.multiLineText)
+
+    actual_error_message = SyncConnection.get_permission_error_message()
+
+    # wait for error message to disappear
+    squish.waitFor(
+            lambda: SyncConnection.get_permission_error_message() == "",
+            get_config("minSyncTimeout") * 1000,
+        )
```
saw-jan: we can have a separate method in the page object file
```suggestion
    def wait_for_error_label(to_exist=True):
        status = squish.waitFor(
            lambda: object.exists(SyncConnection.PERMISSION_ERROR_LABEL) == to_exist,
            SyncConnection.PERMISSION_ERROR_WAIT_TIMEOUT * 1000
        )
        if not status:
            <raise error>
```
then we can use it to wait for label to exist or disappear:
`wait_for_error_label(True)`, `wait_for_error_label(False)`

test/gui/shared/scripts/pageObjects/SyncConnection.py
saw-jan: if we add `wait_for_error_label` method then we can directly return like this. this will throw if the label is not found
```suggestion
        return str(squish.waitForObject(SyncConnection.PERMISSION_ERROR_LABEL).text)
```

test/gui/shared/scripts/pageObjects/SyncConnection.py
```diff
@@ -2,8 +2,12 @@
 import squish
 import object  # pylint: disable=redefined-builtin
 
+from helpers.ConfigHelper import get_config
+
 
 class SyncConnection:
+    PERMISSION_ERROR_WAIT_TIMEOUT = 10
```
saw-jan:
```suggestion
    WAIT_ERROR_LABEL_TIMEOUT = 10
```

test/gui/shared/scripts/pageObjects/SyncConnection.py
```diff
@@ -142,3 +151,20 @@ def confirm_folder_sync_connection_removal():
...
+        )
+        if not status:
+            action = "appear" if to_exist else "disappear"
+            raise AssertionError(f"Permission error label did not {action}")
+
+    @staticmethod
+    def get_permission_error_message():
+        """Get the permission error message text"""
+        SyncConnection.wait_for_error_label(True)  # Wait for label to appear
+        return str(squish.waitForObject(SyncConnection.PERMISSION_ERROR_LABEL, get_config("lowestSyncTimeout") * 1000).text)
```
saw-jan: you can keep the default timeout
```suggestion
        return str(squish.waitForObject(SyncConnection.PERMISSION_ERROR_LABEL).text)
```

PR #513, opened 2025-09-01
test: add test for checking status of shares removed by sharee

anon-pradip:
This PR adds a test coverage for checking if the resources deleted in the server by a `sharee` with `editor` permission role are also deleted in the local file system:
```gherkin
Scenario: Sharee removes the shared resources
```

### DIFF
--- a/test/gui/tst_spaces/test.feature
+++ b/test/gui/tst_spaces/test.feature
@@ -66,3 +66,18 @@ Feature: Project spaces
         When the user removes the folder sync connection
         Then the sync folder list should be empty
         But the file "testfile.txt" should exist on the file system
+
+
+    Scenario: Sharee removes the shared resources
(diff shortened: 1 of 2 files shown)


test/gui/tst_spaces/test.feature
```diff
@@ -81,3 +81,17 @@ Feature: Project spaces
         Then the following activities should be displayed in not synced table
             | resource      | status      | account                              |
             | simple-folder | Blacklisted | Alice Hansen@%local_server_hostname% |
+
+    @issue-435
+    Scenario: Sharee removes the shared resources
```
prashant-gurung899:
```suggestion
    Scenario: Sharee with Editor role deletes the shared resource 
```

prashant-gurung899 (approved):

PR #518, opened 2025-09-04
set new squish license

individual-it:


kulmann (approved):

individual-it:
currently blocked as the firewall rules are still not correct

TheOneRing:
Irrelevant since we moved away from squish.

PR #519, opened 2025-09-04
test: selective sync for selected folders

anon-pradip:


test/gui/shared/steps/file_context.py
```diff
@@ -37,6 +37,17 @@ def file_exists(file_path, timeout=1000):
...
+    return squish.waitFor(
+        lambda: not isdir(sanitize_path(folder_path)),
+        timeout,
+    )
+
+def file_not_exists(file_path, timeout=1000):
+    return squish.waitFor(
+        lambda: not isfile(sanitize_path(file_path)),
+        timeout,
+    )
```
individual-it: I don't think this will work.
If the file/folder does not exist at the start but appears for some reasons a bit later this functions will not catch it.
the `waitFor` will exit after the first iteration because `not isfile(sanitize_path(file_path))` will be true.
I think you will have to use 
``` resource_exists = file_exists(
                resource_path, get_config('maxSyncTimeout') * 1000
            )
```
and then wait for the timeout. but maybe you can use a shorter timeout.
Same as it was already before in `@Then(r'^the (file|folder) "([^"]*)" should not exist on the file system$', regexp=True)`

test/gui/shared/scripts/pageObjects/SyncConnectionWizard.py
```diff
@@ -162,38 +168,7 @@ def deselect_all_remote_folders():
...
-                    squish.mouseClick(
-                        squish.waitForObject(selector),
-                        11,
-                        11,
-                        squish.Qt.NoModifier,
-                        squish.Qt.LeftButton,
-                    )
-                else:
-                    squish.doubleClick(squish.waitForObject(selector))
+        SyncConnectionWizard.select_or_unselect_folders_to_sync(folders, select=True)
```
individual-it: why not calling `SyncConnectionWizard.select_or_unselect_folders_to_sync` directly where now `select_folders_to_sync` is used?

test/gui/shared/scripts/pageObjects/SyncConnectionWizard.py
```diff
@@ -297,3 +272,51 @@ def is_add_sync_folder_button_enabled():
         return squish.waitForObjectExists(
             SyncConnectionWizard.ADD_FOLDER_SYNC_BUTTON
         ).enabled
+
+    @staticmethod
+    def select_or_unselect_folders_to_sync(folders, select=True):
```
individual-it: the naming is confusing as this is not about select and unselect but about two different wizards that have slightly different locators

individual-it (changes requested):

test/gui/shared/scripts/pageObjects/SyncConnectionWizard.py
anon-pradip: Because the step definition mentions `selection` explicitly. So including `select_folders_to_sync` makes it more readable and for `Encapsulation` also.

test/gui/shared/scripts/pageObjects/SyncConnectionWizard.py
anon-pradip: I introduced `in_choose_what_to_sync_dialog` flag to be more specific related to specific widget

test/gui/shared/scripts/pageObjects/SyncConnectionWizard.py
```diff
@@ -297,3 +269,77 @@ def is_add_sync_folder_button_enabled():
...
+                    )
+                else:
+                    squish.doubleClick(squish.waitForObject(selector))
+
+    @staticmethod
+    def confirm_choose_what_to_sync_selection():
+        squish.clickButton(squish.waitForObject(names.stackedWidget_OK_QPushButton))
+
+    @staticmethod
+    def _handle_folder_selection(folders, should_select, in_choose_what_to_sync_dialog):
```
saw-jan:
```suggestion
    def handle_folder_selection(folders, should_select, in_choose_what_to_sync_dialog):
    # or
    def __handle_folder_selection(folders, should_select, in_choose_what_to_sync_dialog):
```

test/gui/shared/scripts/pageObjects/SyncConnectionWizard.py
```diff
@@ -297,3 +269,77 @@ def is_add_sync_folder_button_enabled():
...
+        )
+
+    @staticmethod
+    def select_folders_to_sync(folders, in_choose_what_to_sync_dialog=False):
+        SyncConnectionWizard._handle_folder_selection(
+            folders, should_select=True, in_choose_what_to_sync_dialog=in_choose_what_to_sync_dialog
+        )
+
+    @staticmethod
+    def get_folder_tree_locator(in_choose_what_to_sync_dialog=False):
```
saw-jan:
```suggestion
    def get_folder_tree_locator(new_sync_connection_wizard=False):
```

saw-jan (approved):

Issue #435, opened 2025-07-31
Failed to register sync root after re-adding recently removed account connection in Windows OS


anon-pradip:
### Description:
When trying to re-add an account connection that has just been removed then we get the following error in both `Personal` and `Shares` space:
`Failed to register sync root WindowsError - 7ff8fffb;  Access is denied.`

---
### Steps to reproduce:
1. Add an account connection, let's say `admin`
2. Remove account connection: `Manage Account` &rarr; `Remove`
3. Add the account which was just removed (`admin`)

> [!NOTE]  
> Version: `Version 3.0.0-git`
> OpenCloud 3.0.0-git [#683a43
](https://github.com/opencloud-eu/desktop/commit/683a43a7c86bce8a94a493ec8e44d38878803838)
### Screenshots/Videos (If any):

https://github.com/user-attachments/assets/051a5754-c8eb-4482-9b03-37ff7926d6f7

### More info:
I think the client is failing to delete some files like: `.sync_journal.db` from sync root path

anon-pradip:
cc @TheOneRing

TheOneRing:
I wasn't able to reproduce the issue, does it occur every time?

anon-pradip:
> I wasn't able to reproduce the issue, does it occur every time?

Desktop client built with the commit (latest as of today) https://github.com/opencloud-eu/desktop/commit/7b9c9a6d5ae39ecafd21aa7785fa5c3dffb2541b, still the same issue exists. Please perform those actions in `Steps to reproduce` a bit faster.:

https://github.com/user-attachments/assets/eca5b6c0-5a6f-4c0b-ba6d-a34faebda78d

anon-pradip:
> Is there any chance that `~/OpenCloud (2)` is a leftover of a different test?
FYKI, This error is coming not only on tests but when I run the built `opencloud.exe ` and add an account and remove that account and again add that account, this error comes.

Versions:
```
OpenCloud 3.0.0-git [e3a8ca](https://github.com/opencloud-eu/desktop/commit/e3a8ca1241283924c7547a640a7b159519fa9db0)
Libraries Qt 6.8.3, OpenSSL 3.4.1 11 Feb 2025
Using virtual files plugin: cfapi
OS: windows-10.0.22621 (build arch: x86_64, CPU arch: x86_64)
QPA: windows
```

Video:

https://github.com/user-attachments/assets/797b3a5d-5c01-4a49-a998-7b28a382428e

[...]

PR #530, opened 2025-09-12
Test: Refactor file move operations and add 50MB file sync test

anon-pradip:
## Description
**`desktop/test/gui/shared/steps/file_context.py`:**
- **Added**: New step definition:
  - `@Given` step for creating files with size in sync folder
- Removed the unused `When` step:
    ```gherkin
     @When(
       r'user "([^"]*)" moves (?:folder|file) "([^"]*)" to the temp folder',
       regexp=True,
     )

---
Part of:
#370 

---

Blockage: Could not test on windows due to https://github.com/opencloud-eu/desktop/issues/435


test/gui/shared/steps/file_context.py
```diff
@@ -298,32 +315,15 @@ def step(context, file_number, file_size, folder_name):
...
-
-@When(
-    r'user "([^"]*)" moves (?:folder|file) "([^"]*)" to the temp folder',
-    regexp=True,
-)
-def step(context, _, resource_name):
-    source_dir = get_resource_path(resource_name)
-    destination_dir = get_temp_resource_path(resource_name)
-    shutil.move(source_dir, destination_dir)
+    move_resource(username, resource_name, '/', True)
```
individual-it: I think that new function, even if it saves some line of code, makes the code less readable

test/gui/tst_moveFilesFolders/test.feature
```diff
@@ -65,3 +65,12 @@ Feature: move file and folder
...
         And as "Alice" file "folder1/file1.txt" should not exist in the server
         And as "Alice" file "folder1/file2.txt" should not exist in the server
+
+    @skipOnWindows @issue-435
+	Scenario: Syncing a 50MB file moved into the local sync folder
+        Given user "Alice" has set up a client with default settings
+        And user "Alice" has created a folder "Folder1" inside the sync folder
+        And user "Alice" has created a file "newfile.txt" with size "50MB" in the temp folder
+        When user "Alice" moves file "newfile.txt" from the temp folder to "Folder1" in the sync folder
+        And the user waits for file "Folder1/newfile.txt" to be synced
```
individual-it: I think it will not make any real difference whether we create the file inside the sync folder or inside the temp folder and then move it.
So we can  just simply create it directly in the sync folder and that will save us the 2 new steps and reduce complexity.
Also the whole feature is about moving files / folders within the sync folder. Moving files from outside of the sync folder to inside is basically same as creating them freshly

individual-it (changes requested):

test/gui/shared/steps/file_context.py
anon-pradip: Then previous code should be fine

prashant-gurung899:
Please update the description.
Other LGTM :+1:

prashant-gurung899 (approved):

PR #537, opened 2025-09-15
Test: move resources from different sub-levels to sync root

anon-pradip:
Added a scenario for moving file and folder from different sub-levels to the root of the sync folder:

```gherkin
Scenario: Move resources from different sub-levels to sync root
```

---

part of: #370 

---

Blockage: Could not test on windows due to https://github.com/opencloud-eu/desktop/issues/435


prashant-gurung899:
LGTM :+1:

PR #542, opened 2025-09-17
Test: add test for multibyte characters in filename

anon-pradip:
This PR adds a scenario for testing if the multi-byte characters in the resource name are read properly or not:
```gherkin
    Scenario Outline: File with long multi-byte characters name can be synced (76 characters, 255 bytes including extension)
        Given user "Alice" has set up a client with default settings
        When user "Alice" creates a file "<filename>" with the following content inside the sync folder
            """
            test content
            """
        And the user waits for file "<filename>" to be synced
        Then as "Alice" file "<filename>" should exist in the server
        Examples:
            | filename                                                                                                                                                  |
            | 𒁰𒁱𒁲𒁳𒁴𒁵𒁶𒁷𒁸𒁹𒁺𒁻𒁼𒁾𒁿𒁰𒁱𒁲𒁳𒁴𒁵𒁶𒁷𒁸𒁹𒁺𒁻𒁼𒁾𒁿𒁰𒁱𒁲𒁳𒁴𒁵𒁶𒁷𒁸𒁹𒁺abôǣฎพฒฆ๘ตกกผพฒณญไใๅำ๊๒๔๗๘รศฬอฮ.txt |

```

The characters seems like empty box in GH preview but exact format can be found when you view after clicking `Edit` here in github. Also I am sharing screenshot to illustrate the exact format used in the test:
image

Locally the test passes, lets see in CI

### DIFF
--- a/test/gui/tst_syncing/test.feature
+++ b/test/gui/tst_syncing/test.feature
@@ -525,3 +525,16 @@ Feature: Syncing files
             | resource                 | status      | account                              |
             | simple-folder/sub-folder | Blacklisted | Brian Murphy@%local_server_hostname% |
             | simple-folder/simple.pdf | Blacklisted | Brian Murphy@%local_server_hostname% |
+
+    @skipOnWindows @issue-435
+    Scenario Outline: File with long multi-byte characters name can be synced (76 characters, 255 bytes including extension)
+        Given user "Alice" has set up a client with default settings
+        When user "Alice" creates a file "<filename>" with the following content inside the sync folder
+            """
+            test content
+            """
+        And the user waits for file "<filename>" to be synced
+        Then as "Alice" file "<filename>" should exist in the server
+        Examples:
+            | filename                                                                    |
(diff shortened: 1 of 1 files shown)


test/gui/tst_syncing/test.feature
```diff
@@ -525,3 +525,16 @@ Feature: Syncing files
...
+    Scenario Outline: File with long multi-byte characters name can be synced (76 characters, 255 bytes including extension)
+        Given user "Alice" has set up a client with default settings
+        When user "Alice" creates a file "<filename>" with the following content inside the sync folder
+            """
+            test content
+            """
+        And the user waits for file "<filename>" to be synced
+        Then as "Alice" file "<filename>" should exist in the server
+        Examples:
+            | filename                                                                    |
```
anon-pradip, line 539: The formatting issue here is due to special characters. Formatting is correct in IDE

individual-it (approved):

individual-it:
so the max length is based on count of bytes not characters?

anon-pradip:
> so the max length is based on count of bytes not characters?

Yes. Which means the number of characters should be accommodated within these 255 bytes and number of bytes will vary depending on the character encoding used.

PR #555, opened 2025-09-25
WIP: Add XAttr based plugin for VFS for Linux and later Mac

dragotin:


src/gui/socketapi/socketapi.cpp
```diff
@@ -547,7 +568,7 @@ void SocketApi::command_MAKE_AVAILABLE_LOCALLY(const QString &filesArg, SocketLi
             continue;
 
         // Update the pin state on all items
-        std::ignore = data.folder->vfs().setPinState(data.folderRelativePath, PinState::AlwaysLocal);
+        std::ignore = data.folder->vfs().setPinState(data.localPath, PinState::AlwaysLocal);
```
TheOneRing: Hmm why the change?

src/libsync/common/pinstate.cpp
```diff
@@ -19,6 +19,26 @@
 
 using namespace OCC;
 
+
+template <>
+QString Utility::enumToDisplayName(PinState pState)
```
TheOneRing: enumToDisplayName is meant for translated strings which are displayed to the user.
For technical use we should rely on https://github.com/opencloud-eu/desktop/blob/b246df06c49e758778d7fa0d9badc4b0c5757484/src/libsync/common/utility.h#L261 which by default returns the enum value as string.
E.g. `AlwaysLocal`

src/libsync/vfs/vfs.h
```diff
@@ -252,9 +253,9 @@ public Q_SLOTS:
      */
     virtual void startImpl(const VfsSetupParams &params) = 0;
 
-private:
     // the parameters passed to start()
     std::unique_ptr<VfsSetupParams> _setupParams;
```
TheOneRing, line 258: There is already a getter for params, no need to expose members.

src/plugins/vfs/xattr/vfs_xattr.cpp
```diff
@@ -0,0 +1,474 @@
...
+        // owner is set. See if it is us
+        if (owner.value() == xattrOwnerString()) {
+            // all good
+        } else {
+            qCDebug(lcVfsXAttr) << "Root-FS has a different owner" << owner.value() << "Not our vfs!";
+            err = Vfs::tr("The sync path is claimed by a different cloud, please check your setup");
+            return;
+        }
+    }
+    if (err.isEmpty())
```
TheOneRing: In general, the code style is to use {} even with single statements.

src/plugins/vfs/xattr/vfs_xattr.cpp
```diff
@@ -0,0 +1,474 @@
...
+    // Lets claim the sync root directory for us
+    const auto path = FileSystem::toFilesystemPath(params.filesystemPath);
+
+    auto owner = FileSystem::Xattr::getxattr(path, ownerXAttrName);
+    QString err;
+
+    if (!owner) {
+        // set the owner to opencloud to claim it
+        if (!FileSystem::Xattr::setxattr(path, ownerXAttrName, xattrOwnerString() )) {
+            err = Vfs::tr("Unable to claim the sync root for files on demand");
```
TheOneRing:
```suggestion
            err = tr("Unable to claim the sync root for files on demand");
```

src/plugins/vfs/xattr/vfs_xattr.cpp
```diff
@@ -0,0 +1,474 @@
...
+            return;
+        }
+    } else {
+        // owner is set. See if it is us
+        if (owner.value() == xattrOwnerString()) {
```
TheOneRing: Could you remove the else branch by modifying the check

src/plugins/vfs/xattr/vfs_xattr.h
```diff
@@ -0,0 +1,105 @@
...
+#pragma once
+
+#include <QObject>
+#include <QScopedPointer>
+
+#include "vfs/vfs.h"
+#include "common/plugin.h"
+#include "common/result.h"
+
+#include "config.h"
```
TheOneRing: Please don't include config.h

src/plugins/vfs/xattr/vfs_xattr.cpp
```diff
@@ -0,0 +1,474 @@
...
+{
+    return XAttr;
+}
+
+QString VfsXAttr::xattrOwnerString() const
+{
+    auto s = QByteArray(APPLICATION_EXECUTABLE);
+    s.append(":");
+    s.append(_setupParams->account->uuid().toByteArray(QUuid::WithoutBraces));
+    return QString::fromUtf8(s);
```
TheOneRing:
```suggestion
{
 return u"%1:%2"_s.arg(Theme::instance()->appName() ,_setupParams->account->uuid().toString(QUuid::WithoutBraces));
```
Why was it stored as hex?

src/plugins/vfs/xattr/vfs_xattr.cpp
```diff
@@ -0,0 +1,474 @@
...
+{
+    return true;
+}
+
+PlaceHolderAttribs VfsXAttr::placeHolderAttributes(const QString& path)
+{
+    PlaceHolderAttribs attribs;
+    const auto fPath = FileSystem::toFilesystemPath(path);
+
+    // lambda to handle the Optional return val of xattrGet
```
TheOneRing:
```suggestion
```

src/plugins/vfs/xattr/vfs_xattr.cpp
```diff
@@ -0,0 +1,474 @@
...
+    return attribs;
+}
+
+OCC::Result<void, QString> VfsXAttr::addPlaceholderAttribute(const QString &path, const QString& name, const QString& value)
+{
+    if (!name.isEmpty()) {
+        auto success = FileSystem::Xattr::setxattr(FileSystem::toFilesystemPath(path), name, value);
+        // Q_ASSERT(success);
+        if (!success) {
+            return Vfs::tr("Failed to set the extended file attribute");
```
TheOneRing:
```suggestion
            return tr("Failed to set the extended file attribute");
```

src/plugins/vfs/xattr/vfs_xattr.cpp
```diff
@@ -0,0 +1,474 @@
...
+            item->_inode = inode.value();
+        } else {
+            qCWarning(lcVfsXAttr) << u"Failed to get inode for" << targetPath;
+        }
+
+        // set the xattrs
+        // the file is not virtual any more, remove the xattrs. No state xattr means local available data
+        bool ok{true};
+        ok = ok && FileSystem::Xattr::removexattr(targetPath, stateXAttrName);
+        if (!ok) {
```
TheOneRing: Could we get rid of the `ok` by just using a sigle `if(removexattr() && removexattr() &&...`

src/plugins/vfs/xattr/vfs_xattr.cpp
```diff
@@ -0,0 +1,474 @@
...
+        return file.errorString();
+    }
+    file.write("");
+    file.close();
+
+    FileSystem::Xattr::removexattr(FileSystem::toFilesystemPath(path), actionXAttrName); // remove the action xattr
+
+    // FIXME only write attribs if they're different, and/or all together
+    addPlaceholderAttribute(path, fileSizeXAttrName, QString::number(item._size));
+    addPlaceholderAttribute(path, stateXAttrName, QStringLiteral("virtual"));
```
TheOneRing: we should not use a hardcoded string here.

src/plugins/vfs/xattr/vfs_xattr.cpp
```diff
@@ -0,0 +1,474 @@
...
+        qCWarning(lcVfsXAttr) << u"Ignoring hydration request for running hydration for fileId" << fileId;
+        return {};
+    }
+
+    HydrationJob *hydration = new HydrationJob(this, fileId, std::make_unique<QFile>(targetPath), nullptr);
+    hydration->setTargetFile(targetPath);
+    _hydrationJobs.insert(fileId, hydration);
+
+    // set an action attrib
+    addPlaceholderAttribute(targetPath, actionXAttrName, QStringLiteral("hydrate"));
```
TheOneRing: we should not use a hardcoded string here.

src/plugins/vfs/xattr/vfs_xattr.cpp
```diff
@@ -0,0 +1,474 @@
...
+    });
+
+    return hydration;
+}
+
+bool VfsXAttr::needsMetadataUpdate(const SyncFileItem &item)
+{
+    // return true if file exists
+    const auto path = item.localName();
+    QFileInfo fi{path};
```
TheOneRing: QFileInfo::exists() is faster than creating a QFileInfo object.
Also those local boolean variables are not quite our code style.
Just return true/false.

src/plugins/vfs/xattr/vfs_xattr.cpp
```diff
@@ -0,0 +1,474 @@
...
+    }
+    qCDebug(lcVfsXAttr) << "returning" << re;
+    return re;
+}
+
+bool VfsXAttr::isDehydratedPlaceholder(const QString &filePath)
+{
+    const auto fi = QFileInfo(filePath);
+    bool re{false};
+    if (fi.exists()) {
```
TheOneRing: see 314

src/plugins/vfs/xattr/vfs_xattr.cpp
```diff
@@ -0,0 +1,474 @@
...
+        setPinState(systemFileName, PinState::Excluded);
+        return;
+    }
+
+    qCDebug(lcVfsXAttr) << systemFileName << fileStatus;
+}
+
+QString VfsXAttr::pinStateToString(PinState pState) const
+{
+    QString re;
```
TheOneRing: Just return the values, this is not C 89 where we need to declare out variables at the beginning of a function.

src/plugins/vfs/xattr/vfs_xattr.h
```diff
@@ -0,0 +1,105 @@
...
+    QString _action;
+    QString _state;
+
+};
+}
+
+namespace OCC {
+class HydrationJob;
+
+using namespace xattr;
```
TheOneRing: We should not use `using namespace` in a header.

src/gui/socketapi/socketapi.cpp
dragotin: Fixed in 7504eae8612706f96a793873c9b71c5d8ebfaf94, changed the implementation in vfs_xattr to consume relative paths as the interface requires

src/libsync/common/pinstate.cpp
dragotin: Removed `enumToDisplayName()` for now as it is unused anyway.

src/plugins/vfs/xattr/vfs_xattr.cpp
dragotin: random, fixed with your suggestion

src/plugins/vfs/xattr/vfs_xattr.cpp
dragotin: Removed the `ok` flag. That means that we can not guarantee consistency of the meta data if one of the operations fails. But it is highly unlikely that this happens because we created the file just before.

src/libsync/common/pinstate.cpp
```diff
@@ -19,26 +19,6 @@
 
 using namespace OCC;
 
-
-template <>
-QString Utility::enumToDisplayName(PinState pState)
```
TheOneRing: you forgot the declaration in the header

src/plugins/vfs/xattr/vfs_xattr.cpp
```diff
@@ -267,7 +260,7 @@ Result<void, QString> VfsXAttr::createPlaceholder(const SyncFileItem &item)
 
     // FIXME only write attribs if they're different, and/or all together
     addPlaceholderAttribute(path, fileSizeXAttrName, QString::number(item._size));
-    addPlaceholderAttribute(path, stateXAttrName, QStringLiteral("virtual"));
```
TheOneRing: How about just using pinStateToString that would make it more clear?

dragotin: pinState != state, but we can add that in a subsequent PR

PR #564, opened 2025-10-06
Hash server url

TheOneRing:


src/gui/updatenotifier.cpp
```diff
@@ -29,37 +29,47 @@ void UpdateNotifier::checkForUpdates(const AccountPtr &account)
...
-                            const QString buttonText = tr("Open Download Page");
-                            notification.setButtons({buttonText});
-                            auto *sysNotification = ocApp()->systemNotificationManager()->notify(std::move(notification));
-                            connect(sysNotification, &SystemNotification::buttonClicked, this, [buttonText, url](const QString &button) {
-                                if (button == buttonText) {
-                                    QDesktopServices::openUrl(url);
+                                    const QString buttonText = tr("Open Download Page");
+                                    notification.setButtons({buttonText});
+                                    auto *sysNotification = ocApp()->systemNotificationManager()->notify(std::move(notification));
+                                    connect(sysNotification, &SystemNotification::buttonClicked, this, [buttonText, url](const QString &button) {
```
kulmann, line 56: how often do I get the notification? 🤔 once per software restart?

src/gui/updatenotifier.cpp
```diff
@@ -29,37 +29,47 @@ void UpdateNotifier::checkForUpdates(const AccountPtr &account)
...
+                            }
+                        } else {
+                            qCWarning(lcUpdateNotifier) << u"Update check failed with HTTP code" << updateJob->httpStatusCode();
                         }
-                    }
+                    });
+                    updateJob->start();
                 } else {
-                    qCWarning(lcUpdateNotifier) << u"Update check failed with HTTP code" << job->httpStatusCode();
+                    qCWarning(lcUpdateNotifier) << u"Update check config.json with HTTP code" << job->httpStatusCode();
```
kulmann, line 69: are the warnings user facing? because then they'd be very technical 🙈

src/gui/updatenotifier.cpp
TheOneRing: yes

src/gui/updatenotifier.cpp
TheOneRing: only logs

src/gui/updatenotifier.cpp
kulmann: Is the effort reasonably to show it only once per new available version? There might be users who can't install an update without an admin, so they can't do anything about the notification.

TheOneRing: The system notification is usually only displayed for a couple of seconds, so it's not too intrusive.
If we wanted to implement a "skip this update" functionality, we would need to persist the decision. Also, we would need to add a second action to the notification, which is currently not implemented on Mac.

kulmann: sounds all reasonable :-) thank you!

kulmann (approved):
LGTM 👍

PR #584, opened 2025-10-16
fix UI tests cleanup

individual-it:
There have been at least multiple issues why the client config was not correctly cleaned up after the tests:

1. the accounts on the server have been deleted before the client cleanup, causing the client to be stuck in the login window
2. `and child_widget.objectName` caused the code to never go into that path, so the cleanup was never attempted
3. `get_active_widget()`  has also to ignore widgets where the name is empty

---

Fix `PermissionError` in Re-add Account Test on Windows

Fixes `WinError 32` when removing account connection by adding synchronization 
before filesystem cleanup.

Changes:
- Add wait_until_account_is_removed method to ensure account is fully removed from UI before deleting sync folder
- Update remove_connection_for_user step to wait for account removal before calling `shutil.rmtree()`

This prevents race condition where sync journal DB files are still locked by the application during cleanup.

### DIFF
--- a/test/gui/shared/scripts/pageObjects/AccountSetting.py
+++ b/test/gui/shared/scripts/pageObjects/AccountSetting.py
@@ -166,3 +166,21 @@ class AccountSetting:
         displayname = substitute_inline_codes(displayname)
         Toolbar.open_account(displayname)
         AccountSetting.remove_account_connection()
+
+    @staticmethod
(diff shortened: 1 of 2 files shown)


individual-it:
@anon-pradip please review again I found an other issue and fixed it

individual-it (commented):
LGTM

PR #701, opened 2025-11-18
Fix canonicalPath for drives (c:)

TheOneRing:


dragotin (approved):
This one is hard. It has new tests, that is good, so it is consitent. If the behaviour changed compared to before is a different story. Let's test.

PR #711, opened 2025-11-24
Update SVG logo with new design

db-ot:
added new centered polygon to svg in 000


flimmy:
before the change:
pause_stateno_state

PR #726, opened 2025-11-27
Clear selective sync with windows vfs

TheOneRing:


src/gui/folder.cpp
```diff
@@ -543,6 +543,12 @@ void Folder::startVfs()
         _vfs->fileStatusChanged(stateDbFile + QStringLiteral("-wal"), SyncFileStatus::StatusExcluded);
         _vfs->fileStatusChanged(stateDbFile + QStringLiteral("-shm"), SyncFileStatus::StatusExcluded);
         _engine->setSyncOptions(loadSyncOptions());
+
+        if (_vfs->mode() == Vfs::WindowsCfApi) {
```
dragotin, line 547: shouldn't that rather be `if (_vfs->mode() != Vfs::Off) {` ?

Otherwise we start over again with the other vfs'ses

PR #735, opened 2025-12-01
Fix high cpu usage during cfapi hydration

TheOneRing:
cfapi use a bigger buffer for hydrations the call to CfExecute is quite expensive.

### DIFF
--- a/src/plugins/vfs/cfapi/hydrationdevice.cpp
+++ b/src/plugins/vfs/cfapi/hydrationdevice.cpp
@@ -15,7 +15,7 @@ using namespace OCC::FileSystem::SizeLiterals;
 Q_LOGGING_CATEGORY(lcCfApiHydrationDevice, "sync.vfs.cfapi.hydrationdevice", QtDebugMsg)
 namespace {
 constexpr auto ChunkSize = 4_KiB;
-constexpr auto BufferSize = ChunkSize * 10;
+constexpr auto BufferSize = 4_MiB;
 }
 
 
@@ -70,6 +70,7 @@ CfApiWrapper::HydrationDevice::HydrationDevice(const CfApiWrapper::CallBackConte
     , _context(context)
     , _totalSize(totalSize)
 {
+    // we reserve a fixed size and don't expect to ever grow the array
     _buffer.reserve(BufferSize);
 }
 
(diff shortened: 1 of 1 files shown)


src/plugins/vfs/cfapi/hydrationdevice.cpp
```diff
@@ -81,14 +81,15 @@ qint64 CfApiWrapper::HydrationDevice::readData(char *, qint64)
...
     _buffer.append(data, len);
-    // the buffer should not grow above BufferSize
-    Q_ASSERT(_buffer.size() <= BufferSize);
     const bool isLastChunk = _offset + _buffer.size() == _totalSize;
-
-    if (_buffer.size() >= ChunkSize || isLastChunk) {
+    // only write if the buffer is decently filled, or we are in the last chunk
+    if (_buffer.size() >= BufferSize * 0.9 || isLastChunk) {
+        // the buffer should not grow above BufferSize
+        Q_ASSERT(_buffer.size() <= BufferSize);
```
aduffeck, line 89: What was the reason for moving that assertion down here? If its purpose is to ensure that the buffer never ever exceeds `BufferSize` I guess it would be better to leave it where it was before, no?

src/plugins/vfs/cfapi/hydrationdevice.cpp
```diff
@@ -81,14 +81,15 @@ qint64 CfApiWrapper::HydrationDevice::readData(char *, qint64)
 qint64 CfApiWrapper::HydrationDevice::writeData(const char *data, qint64 len)
 {
     _buffer.append(data, len);
-    // the buffer should not grow above BufferSize
-    Q_ASSERT(_buffer.size() <= BufferSize);
     const bool isLastChunk = _offset + _buffer.size() == _totalSize;
```
aduffeck: I'm just wondering... Are there (unexpected) cases where the size could be bigger than `_totalSize`? Maybe changing this to

```suggestion
    const bool isLastChunk = _offset + _buffer.size() >= _totalSize;
```

could help to prevent issues.

src/plugins/vfs/cfapi/hydrationdevice.cpp
TheOneRing: There is not really a reason to test on every single call. It basically should never happen and checking it here should work as well?

src/plugins/vfs/cfapi/hydrationdevice.cpp
TheOneRing: hmm this should never happen, but changing the check makes sense thx.

src/plugins/vfs/cfapi/hydrationdevice.cpp
aduffeck: fair enough

PR #766, opened 2025-12-21
feat(oauth): support desktop-specific OIDC issuer and `client_id` discovery

### LINKED ISSUE #246: Support different OIDC issuer for desktop
## Problem 
Both web and desktop clients share one issuer (`OC_OIDC_ISSUER`) but have different client id and redirect URI. 
So the desktop app (which has its own client ID and redirect URI) fails authentication as it uses

## Steps to Reproduce  
1. In Authentik, create two apps:  
   - **web** (ID: `web`, issuer `https://auth.domain.com/application/o/opencloud/`, redirect `regex: ^https:\/\/opencloud\.domain\.com\/.*$`)  
   - **desktop** (ID: `OpenCloudDesktop`, issuer `https://auth.domain.com/application/o/opencloud-desktop/`, redirect `regex: ^http:\/\/(127\.0\.0\.1|localhost).*$`)  
2. Set `OC_OIDC_ISSUER` to the web URL and `OC_OIDC_CLIENT_ID=web`.  
3. Set `OC_OIDC_ISSUER=https://auth.domain.com/application/o/opencloud/`.
4. Launch the desktop app → it still uses the web issuer and client ID, and auth fails.
    - uses `https://auth.domain.com/application/o/opencloud/` instead of `https://auth.domain.com/application/o/opencloud-desktop/` (expected since there is no way to configure that)

## Expected
Add support for per-client overrides (e.g. `DESKTOP_OIDC_ISSUER` and `DESKTOP_OIDC_CLIENT_ID`) on the serveer so each app can point to its own OIDC endpoint if possible. 

Or allow to specify the issuer directly in the desktop app instead of using the default web one.

## Environment
Authentik for OIDC, Caddy reverse proxy, OpenCloud (2.0.2), OpenCloud desktop 1.0.0

## Workaroud
As for now, we can simply use `WEB_OIDC_CLIENT_ID=OpenCloudDesktop` and use only one issuer and support both web and desktop redirect URI. This allows both the web client and desktop client uses the same issuer.

pat-s:
fix #246

Enables identity providers that require separate OIDC clients per application type (like Authentik, Kanidm, Zitadel) to work with the desktop client.

- Discover desktop-specific OIDC issuer via WebFinger (`http://openid.net/specs/connect/1.0/issuer/desktop`)
- Use server-provided `client_id` from WebFinger link properties when available
- Persist and restore `client_id` across re-authentication

## Server-side changes

To work, this PR requires changes in https://github.com/opencloud-eu/opencloud. 
A complementary PR is available at https://github.com/opencloud-eu/opencloud/pull/2072.

Additionally, the following env vars must be set on the admin side:

```
WEBFINGER_OIDC_ISSUER_DESKTOP: <>
WEBFINGER_OIDC_CLIENT_ID_DESKTOP: <>
```

Here, `WEBFINGER_OIDC_ISSUER_DESKTOP` is usually the root URL of the identity provider and `WEBFINGER_OIDC_CLIENT_ID_DESKTOP` contains the `client_id` of the respective oAuth project.

---

With these changes, I was able to login into the Desktop app with a dedicated oAuth project through Zitadel (v4.7.0). The implementation should be generic and work with other oAuth providers as well.

### DIFF
--- a/src/gui/newwizard/jobs/discoverwebfingerservicejobfactory.cpp
+++ b/src/gui/newwizard/jobs/discoverwebfingerservicejobfactory.cpp
@@ -21,6 +21,7 @@
 #include <QJsonObject>
 #include <QJsonParseError>
 #include <QNetworkReply>
+#include <QUrl>
 
 namespace OCC::Wizard::Jobs {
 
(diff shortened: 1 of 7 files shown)


guruz:
Hi @pat-s I'm trying to get an overview here about the different open PRs..
Do I see it correctly that this PR is just for client_id, while the https://github.com/opencloud-eu/desktop/pull/847 (+ base PRs or whateber?) is also for scope? 
CC @kaivol

kaivol:
As far as I can tell the implementations differ in some places. 

For example, I would suggest that persisting the `clientId` (see [here](https://github.com/opencloud-eu/desktop/pull/766/changes#diff-030afc7423b8921f965c6ffa0af5f7312a2e7aa06e1046628ddbde8e6394ac50R506)) is not a good idea. 
Also, I'd argue that the additional restructuring in my PRs makes a lot of sense, in particular the improvement described in  https://github.com/opencloud-eu/desktop/pull/776#issue-3802192211.

guruz:
Thank you for the contribution @pat-s ... 

Now https://github.com/opencloud-eu/desktop/pull/847 is merged which also supports customized ```scope``` values so this PR here is not needed anymore.

It would still be great to get testing help if https://github.com/opencloud-eu/desktop/pull/847 works for your setup.

pat-s:
@guruz I'd be happy if somebody from the team would rebase this and get it in - or any other fresh/new PR. Contributing here wasn't a great experience so far and at some point I have mainly given up pushing.

I would still be interested in using OpenCloud one day - but first this problem must be resolved.

TheOneRing:
Closing now that https://github.com/opencloud-eu/desktop/pull/847 is merged

guruz:
Hi all, Testing of pre-release with webfinger client_id/scope support would be appreciated.
It's installable seperately from the main version, so there should be no collision.
https://github.com/opencloud-eu/desktop/releases/tag/v4.0.0-rc.1

PR #779, opened 2026-01-16
test(gui): use current time if timestamp not found

saw-jan:
Use the current time if the timestamp pattern is not found

Fixes:
```
Script Error: could not convert string to float: 'N/A'
```
- https://ci.opencloud.rocks/repos/11/pipeline/340/46
- https://s3.ci.opencloud.eu/public/desktop/340/guiReportUpload/index.html


anon-pradip (approved):
🚀

Issue #787, opened 2026-01-26
Client crashes while adding sync connection after deselecting all remote folders


anon-pradip:
## Steps to reproduce
1. Create a couple of files and folders in server
2. Add account  to desktop client with `Configure synchronization manually` enabled.
    <img width="799" height="231" alt="Image" src="https://github.com/user-attachments/assets/6c571503-b775-42cc-a387-6376da7719c9" />
3. In `Deselect all remote folders` setup window, deselect all remote folders ⚠️    
    <img width="782" height="215" alt="Image" src="https://github.com/user-attachments/assets/a832c67c-dfe8-4174-8c28-3331dd2fd85b" />

> [!IMPORTANT]
> This issue occurs with the client build built with `cmake` from client repo but does not occur with the build built using `openbuild`

5. Click `Add Space` 💥 
### Expected Behavior
The client should remain open, allow selection of the "Personal" space, and continue with the sync setup without crashing.

### Failing CI:
https://ci.opencloud.rocks/repos/11/pipeline/394/49#L1276

### Crash Logs:
<details>
<summary>OpenCloud crash log (first run)</summary>

```
OpenCloud 3.1.0 5556ba Libraries Qt 6.9.2, OpenSSL 3.5.4 30 Sep 2025 Using virtual files plugin: xattr OS: ubuntu-6.8.0-90-generic (build arch: x86_64, CPU arch: x86_64) QPA: xcb OpenCloud
26-01-26 10:34:26:760 [ info gui.account.state ]:	checkConnectivity blocking: false "Admin@host.docker.internal:9200"
26-01-26 10:34:26:760 [ info sync.connectionvalidator ]:	No system proxy set by OS
26-01-26 10:34:26:766 [ info sync.checkserverjob ]:	status.php returns:  QJsonDocument({"channel":"","edition":"rolling","installed":true,"maintenance":false,"needsDbUpgrade":false,"product":"OpenCloud","productname":"OpenCloud","productversion":"4.1.0+e372ae0c","version":"0.1.0.0","versionstring":"0.1.0"})   QNetworkReply::NoError  Reply:  QNetworkReplyHttpImpl(0x594d189c8a60)
26-01-26 10:34:26:766 [ info sync.connectionvalidator ]:	** Application: OpenCloud found:  QUrl("https://host.docker.internal:9200")  with version  "4.1.0+e372ae0c"
[...]

PR #788, opened 2026-01-26
Use xattr backend by default

TheOneRing:
Fixes: #787

### DIFF
--- a/src/gui/folderdefinition.cpp
+++ b/src/gui/folderdefinition.cpp
@@ -104,14 +104,13 @@ FolderDefinition FolderDefinition::load(QSettings &settings)
     folder.virtualFilesMode = Vfs::Off;
 
     QString vfsModeString = settings.value("virtualFilesMode").toString();
-#ifdef Q_OS_WIN
-    // we always use vfs on windows if available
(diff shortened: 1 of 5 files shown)


src/gui/folderdefinition.cpp
```diff
@@ -104,14 +104,13 @@ FolderDefinition FolderDefinition::load(QSettings &settings)
...
+
+    const auto vfs = Utility::isWindows() ? Vfs::WindowsCfApi : Vfs::XAttr;
+    if (auto result = Vfs::checkAvailability(folder.localPath(), vfs); result) {
+        vfsModeString = Utility::enumToString(vfs);
     } else {
-        qCWarning(lcFolder) << u"Failed to upgrade" << folder.localPath() << u"to" << Vfs::WindowsCfApi << result.error();
+        qCWarning(lcFolder) << u"Failed to upgrade" << folder.localPath() << u"to" << vfs << result.error();
     }
-#endif
     if (!vfsModeString.isEmpty()) {
```
dragotin, line 114: Hm, I believe the usage of `vfsModeString` is somehow back and forth conversion...

I think in this code block starting here could be simplified by using `vfs` directly. Not really part of this PR tough.

PR #795, opened 2026-02-04
Mount and unmount openvfs

TheOneRing:
I could split the changes here in two reviews if needed.
The refactoring of the vfs cases to make `checkAvailability` available without an instance of a backend and the mount/unmount part.

I'm currently not happy with the name `checkAvailability` as it does more than that and takes care of the unmounting of the stale mounts.

Depends on: https://github.com/dragotin/openvfs/pull/22

### DIFF
--- a/src/libsync/xattr.cpp
+++ b/src/libsync/xattr.cpp
@@ -9,6 +9,15 @@
 
 namespace OCC {
 namespace FileSystem {
+    bool Xattr::supportsxattr(const std::filesystem::path &path)
+    {
+#ifdef Q_OS_MAC
+        return !(::listxattr(path.c_str(), nullptr, 0, XATTR_NOFOLLOW) == -1 && errno == ENOTSUP);
+#else
+        return !(::listxattr(path.c_str(), nullptr, 0) == -1 && errno == ENOTSUP);
+#endif
+    }
+
     std::optional<QString> Xattr::getxattr(const std::filesystem::path &path, const QString &name)
     {
         QByteArray value;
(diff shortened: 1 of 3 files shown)


src/gui/folder.cpp
```diff
@@ -186,7 +184,12 @@ bool Folder::checkLocalPath()
                 error = pathLengthCheck.error();
             }
 
-            if (error.isEmpty()) {
+            // Potentially upgrade suffix vfs to windows vfs
```
dragotin: This comment does not really make sense any more, right?

src/gui/folder.cpp
```diff
@@ -186,7 +184,12 @@ bool Folder::checkLocalPath()
...
+            OC_ENFORCE(_vfs);
+            const auto result = VfsPluginManager::instance().checkAvailability(path(), _vfs->mode());
+            if (!result) {
+                error = result.error();
+            } else if (error.isEmpty()) {
```
dragotin, line 190: Hm, I think it would be nicer to have the vfs preperation in its own method rather than stuffing it into `checkLocalPath()`. How about `prepVfs()` or something, just to make the purpose more clear?

src/libsync/vfs/vfs.h
```diff
@@ -182,7 +180,7 @@ class OPENCLOUD_SYNC_EXPORT Vfs : public QObject
      * Usually backed by the db's effectivePinState() function but some vfs
      * plugins will override it to retrieve the state from elsewhere.
      *
-     * relFilePath is relative to the sync folder. Can be "" for root folder.
+     * relFilePath is relative to the sync folder. Can be "" for root folder.<
```
dragotin: Leftover char `<`

PR #813, opened 2026-02-18
test(gui): fix VFS test scenarios

saw-jan:
- Fix copying file to the same name
- Add missing step definition
- Rename steps
- Extend vfs test scenario

### DIFF
--- a/test/gui/shared/steps/file_context.py
+++ b/test/gui/shared/steps/file_context.py
@@ -9,7 +9,7 @@ import squish
 
 from helpers.SetupClientHelper import get_resource_path, get_temp_resource_path
 from helpers.SyncHelper import wait_for_client_to_be_ready, listen_sync_status_for_item
-from helpers.ConfigHelper import get_config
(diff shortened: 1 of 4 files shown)


anon-pradip (approved):

test/gui/shared/steps/file_context.py
```diff
@@ -97,26 +97,27 @@ def extract_zip(zip_file_path, destination_dir):
 
 
 def add_copy_suffix(resource_path, resource_type):
+    suffix = ' (Copy)'
```
prashant-gurung899, line 100: When i copied a folder in the same directory then the following appeared:

Image


So, i think the previous code should be fine ❓

prashant-gurung899: Nevermind.

saw-jan: we are just copying current file to a new name, so whatever name will work. Also, I am moving focus to linux vfs (soon to be added but important)

PR #814, opened 2026-02-18
OpenVFS fix pin state

TheOneRing:
Ci as faling as we can't run fuse tests in docker

### DIFF
--- a/src/plugins/vfs/xattr/vfs_xattr.cpp
+++ b/src/plugins/vfs/xattr/vfs_xattr.cpp
@@ -202,10 +202,7 @@ void VfsXAttr::startImpl(const VfsSetupParams &params)
     // merging the channels and piping the output to our log lead to deadlocks
     _openVfsProcess->setProcessChannelMode(QProcess::ForwardedChannels);
     const auto logPrefix = [path = params.root().toString(), this] { return u"[%1 %2] "_s.arg(QString::number(_openVfsProcess->processId()), path); };
-    connect(_openVfsProcess, &QProcess::finished, this, [logPrefix, this] {
-        qCFatal(lcVfsXAttr) << logPrefix() << "finished" << _openVfsProcess->exitCode();
-        _openVfsProcess->deleteLater();
-    });
+    connect(_openVfsProcess, &QProcess::finished, this, [logPrefix, this] { qCFatal(lcVfsXAttr) << logPrefix() << "finished" << _openVfsProcess->exitCode(); });
     connect(_openVfsProcess, &QProcess::started, this, [logPrefix, this] {
         qCInfo(lcVfsXAttr) << logPrefix() << u"started";
         // TODO:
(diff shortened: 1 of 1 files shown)


src/plugins/vfs/xattr/vfs_xattr.cpp
```diff
@@ -190,31 +190,38 @@ void VfsXAttr::startImpl(const VfsSetupParams &params)
...
+        qCInfo(lcVfsXAttr) << logPrefix() << "finished" << _openVfsProcess->exitCode();
+        _openVfsProcess->deleteLater();
     });
-    connect(vfsProcess, &QProcess::started, this, [logPrefix, this] {
+    connect(_openVfsProcess, &QProcess::started, this, [logPrefix, this] {
         qCInfo(lcVfsXAttr) << logPrefix() << u"started";
-        Q_EMIT started();
+        // TODO:
+        // give it time to mount
+        QTimer::singleShot(1s, this, &Vfs::started);
```
dragotin, line 210: Hm, have you thought about using the `started()` signal of QProcess?
Mount is still not established when it comes?

src/plugins/vfs/xattr/vfs_xattr.cpp
```diff
@@ -496,12 +503,17 @@ bool VfsXAttr::setPinState(const QString &folderPath, PinState state)
...
-        return {};
+    for (auto relativePath = FileSystem::Path::relative(folderPath).get();; relativePath = relativePath.parent_path()) {
+        const auto attributes = placeHolderAttributes(params().root() / relativePath);
+        if (!attributes) {
+            qCDebug(lcVfsXAttr) << "Couldn't find pin state for placeholder file" << folderPath;
+            return {};
+        }
+        // if the state is inherited and we still have a parent path, retreive that instead.
+        if (attributes.pinState != OpenVfsConstants::PinStates::Inherited || !relativePath.has_relative_path()) {
+            return convertPinState(attributes.pinState);
```
dragotin, line 515: what if there is no relative path any more (we are at root), but the state is still `Inherited`? Shouldn't that result in something defined?

test/testutils/syncenginetestutils.cpp
```diff
@@ -865,13 +865,14 @@ FakeFolder::FakeFolder(const FileInfo &fileTemplate, OCC::Vfs::Mode vfsMode, boo
         Q_ASSERT(vfs);
     }
 
-    // Ensure we have a valid Vfs instance "running"
-    switchToVfs(vfs);
-
     if (vfsMode != OCC::Vfs::Mode::Off) {
-        const auto pinState = filesAreDehydrated ? OCC::PinState::OnlineOnly : OCC::PinState::AlwaysLocal;
-        OC_ENFORCE(vfs->setPinState(QString(), pinState));
+        connect(vfs.data(), &OCC::Vfs::started, this, [vfs, filesAreDehydrated] {
```
dragotin: Can you add a comment why this is needed?

src/plugins/vfs/xattr/vfs_xattr.cpp
TheOneRing: The signal is emitted once the process started, but that doesn't guarantee that the fuse layer is already initialised.

src/plugins/vfs/xattr/vfs_xattr.cpp
TheOneRing: The value of the root must be defined, we should ensure that either openvfs defaults to something or we set something.

PR #816, opened 2026-02-19
ci: separate cache locations

saw-jan:
Sometimes ci failed to restore the python cache. This was because we store the python cache in the desktop-build location which will get purged (with older-than 1day).

Here, I have separated the python and desktop-build cache paths and added gui-test workflow dependency to the purge-cache workflow to run it at the very end.

Screenshot from 2026-02-19 10-40-21


.woodpecker/build.yaml
```diff
@@ -46,5 +46,5 @@ steps:
       <<: *minio_environment
     commands:
       - mc alias set s3 $MC_HOST $AWS_ACCESS_KEY_ID $AWS_SECRET_ACCESS_KEY
-      - mc cp -a -r /woodpecker/desktop/build/bin s3/$CACHE_BUCKET/desktop-build/${CI_COMMIT_SHA}/
-      - mc ls --recursive s3/$CACHE_BUCKET/desktop-build
+      - mc cp -a -r /woodpecker/desktop/build/bin s3/$CACHE_BUCKET/desktop/bin-build/${CI_COMMIT_SHA}/
```
prashant-gurung899: Just to understand at first glance.
```suggestion
      - mc cp -a -r /woodpecker/desktop/build/bin s3/$CACHE_BUCKET/desktop/client-build/${CI_COMMIT_SHA}/
```

prashant-gurung899: Other LGTM

saw-jan: :+1: would read nicely

PR #818, opened 2026-02-19
Update OPENCLOUD_COMMITID in .woodpecker.env

flimmy:


v-scharf:
We can close it only because there is no more .woodpecker.env in the code

PR #820, opened 2026-02-23
ci: use latest rolling opencloud server version

saw-jan:
Server the latest rolling version of opencloud server for the tests instead of using commit hash which needs to be updated periodically.
This guarantees that the desktop client is working with the current latest rolling version and no manual effort needed to update the server commit


test/gui/woodpecker/script.sh
```diff
@@ -1,35 +1,6 @@
...
-    if [[ "$opencloud_cache" != "" ]]
-    then
-        echo "$OPENCLOUD_COMMITID doesn't exist in cache."
-        ENV="OPENCLOUD_CACHE_FOUND=false\n"
-    else
-      echo "$OPENCLOUD_COMMITID found in cache."
-      ENV="OPENCLOUD_CACHE_FOUND=true\n"
-    fi
-}
+touch .woodpecker.env
```
prashant-gurung899, line 3: Won't there be a `.woodpecker.env` aleady? or am i just confused? 🤔

saw-jan: nope, I removed it as it doesn't contain any data. only needed during the workflow run

prashant-gurung899: Got it :+1:

prashant-gurung899 (approved):

PR #837, opened 2026-03-03
Abort precheck if fuse is not available

TheOneRing:
Depends on https://github.com/opencloud-eu/desktop/pull/839

### DIFF
--- a/src/plugins/vfs/xattr/vfs_xattr.cpp
+++ b/src/plugins/vfs/xattr/vfs_xattr.cpp
@@ -234,6 +234,21 @@ bool VfsXAttr::socketApiPinStateActionsShown() const
 }
 
 
+bool XattrVfsPluginFactory::checkAvailability() const
+{
+#ifdef Q_OS_LINUX
+    if (!FileSystem::Path("/dev/fuse").exists()) {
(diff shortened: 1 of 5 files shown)


src/plugins/vfs/xattr/vfs_xattr.cpp
```diff
@@ -234,6 +234,21 @@ bool VfsXAttr::socketApiPinStateActionsShown() const
...
+{
+#ifdef Q_OS_LINUX
+    if (!FileSystem::Path("/dev/fuse").exists()) {
+        qCWarning(lcVfsXAttr) << u"Fuse is not installed or available on the system";
+        return false;
+    }
+    if (QStandardPaths::findExecutable(u"fusermount3"_s).isEmpty()) {
+        qCWarning(lcVfsXAttr) << u"fusermount3 is not installed on the system";
+        return false;
+    }
```
dragotin, line 247: Maybe add another check if the `openvfsfuse` binary is available and executable?

TheOneRing: Hmm the two here are system checks.
Fuse might not be available on the system.
openvfs is a mandatory dependency of opencloud.
We link the openvfs lib, so the plugin loading would already fail. If the executable is missing.... we would display an error later in the ui.

PR #851, opened 2026-03-22
Sync folder setup

sh4tteredd:


TheOneRing:
We decided a while ago that we don't want to support syncing arbitrary folders as it makes this for normal users way too complicated, besides it also introduces a bunch of corner cases.

PR #858, opened 2026-03-31
Add per-account mTLS client certificate support (PKCS#12)

paolostivanin:
Adds per-account mTLS client certificate support.

- Configure a client certificate for mutual TLS authentication on a **per-account** basis. Certificates are imported from PKCS#12 (`.p12`/`.pfx`) files and persisted in the account settings.
- The certificate can be selected **during account setup** (on the server-URL page), so a server that gates access behind mTLS can be reached for the initial `status.php` / WebFinger / OAuth probes before the account exists — as well as **per account after setup**, via the account context menu.
- CA certificates bundled in the PKCS#12 are trusted for server-chain validation.

Mirrors the Android per-account mTLS work in https://github.com/opencloud-eu/android/pull/116 .

> [!NOTE]
> The network-switch recovery fix that was previously bundled here has been split into its own PR: #969.

Tested on openSUSE Tumbleweed x86_64.


TheOneRing:
Could you please open a separate pr for b261cfdda58370daa00091b3ce646a62add20d63

paolostivanin:
All done!

- This PR is now scoped to per-account mTLS client certificate support only. It additionally adds **login-time selection** (in the setup wizard) so servers that gate access behind mTLS can be onboarded before the account exists.
- The network-switch recovery fix has been split out into its own PR: #969.

guruz:
Fun fact: Commit from 2017 ... c6f4f446192e5f11df80186e391b636471c15984

PR #863, opened 2026-04-09
build: build qtbase with a11y bridge from source

### LINKED ISSUE #860: [A11y] Screen reader cannot read the contents of desktop client
## Steps to reproduce
1. Open desktop client
2. Open screen reader. E.g: orca
3. Focus the desktop client window
4. Nothing from the screen reader ⚠️ 

#### Findings
The problem seems to be how the client is built. The client is built using the craft build system which uses the qtbase library from the kde cache. The cached qtbase doesn't include the necessary a11y implementations/modules.

The fix would be to build the client with qtbase that has necessary a11y modules. When using craft, we need to build the `libs/qt6/qtbase` from source instead of using from cache. And while build the qtbase, make sure to confirm that `ATSPI` is found and `Accessibility` feature says `yes` 
```
Qt Gui:
  Accessibility .......................... yes
```
check the build:
```
strings ./main/linux-gcc-x86_64/lib/libQt6Gui.so | grep -i QSpiAccessibleBridge
```

## Expected behavior
Screen reader reads the contents of desktop client window

## Actual behavior
Screen reader is silent when focusing the desktop client window

## Setup
- Your operating system: `Ubuntu 24.04`
- Client version:
```
OpenCloud 3.0.3.2073 [72fe8f]
Libraries Qt 6.9.2, OpenSSL 3.5.4 30 Sep 2025
Using virtual files plugin: off
OS: ubuntu-6.8.0-106-generic (build arch: x86_64, CPU arch: x86_64)
QPA: xcb
```

saw-jan:
Partially fixes #860
Do not use qtbase from cache but instead build it from source with a11y and a11y-atspi-bridge features enabled.
> [!IMPORTANT]
> The build system MUST have the necessary libraries installed
> - libatspi2.0-dev
> - libdbus-1-dev

To confirm the features, run the following command:
```bash
strings <path-to>/lib/libQt6Gui.so | grep -i QSpiAccessibleBridge
```

### DIFF
--- a/.craft.ini
+++ b/.craft.ini
@@ -60,6 +60,7 @@ dev-utils/msys.ignored = True
 dev-utils/gtk-doc.ignored = True
 
 #libs/qt6.version = ${Variables:QtVersion}
+libs/qt6/qtbase.featureArguments = -DQT_FEATURE_accessibility=ON -DQT_FEATURE_accessibility_atspi_bridge=ON
 
 [windows-cl-msvc2022-x86_64]
 General/ABI = windows-cl-msvc2022-x86_64


.craft.ini
```diff
@@ -60,6 +60,7 @@ dev-utils/msys.ignored = True
 dev-utils/gtk-doc.ignored = True
 
 #libs/qt6.version = ${Variables:QtVersion}
+libs/qt6/qtbase.featureArguments = -DQT_FEATURE_accessibility=ON -DQT_FEATURE_accessibility_atspi_bridge=ON
```
saw-jan, line 63: how can we do it for linux only?

TheOneRing:
As this makes total sense for all of KDE, I'll try to upstream the change and enable accessibility for the KDE Qt packages.

TheOneRing:
It's now enabled upstream https://invent.kde.org/packaging/craft-blueprints-kde/-/merge_requests/1564

saw-jan:
> It's now enabled upstream https://invent.kde.org/packaging/craft-blueprints-kde/-/merge_requests/1564

THank you @TheOneRing

PR #865, opened 2026-04-14
A typo in the Russian translation has been corrected

KeltorHD:


dragotin:
Thank you for your efforts, but we can not take localisation fixes through Github because that is maintained in Transifex: https://explore.transifex.com/opencloud-eu/opencloud-eu/

Please fix the typo there, thank you!

PR #876, opened 2026-04-21
test(gui): fix and enable add-account scenarios

saw-jan:
Part of #861

- Fixed and enabled scenarios:
  ```feature
  @smoke
  Scenario: Re-add an account
  ```
- added a data table parser
  ```py
  table_raw()
  table_rows()
  table_rows_hash()
  table_hashes()
  ```

PR #878, opened 2026-04-22
test(gui): port remove account connection scenario to appium

prashant-gurung899:
Part of #861

Enable scenario:
```feature
@smoke
Scenario: remove an account connection
```

PR #879, opened 2026-04-23
fix(ui): fix click event for a11y

saw-jan:
Fixes: https://github.com/opencloud-eu/desktop/pull/876#discussion_r3122557495

```qml
Accessible.onPressAction: {...}
```
This allows the a11y tools to perform the click action on the element, same action as done inside `onClicked`

Adjust the tests after this merge:
- #881
- #878


guruz (approved):
I haven't tested this but looks OK to me

src/gui/qml/AccountBar.qml
```diff
@@ -79,6 +79,9 @@ Pane {
                 ToolTip.delay: 500
 
                 Accessible.name: accountState.account.displayNameWithHost
+                Accessible.onPressAction: {
```
TheOneRing, line 82: Why is the account button special, shouldn't all buttons in the top bar require this special handling?
I see why it might be needed for the space view, where we use a custom MouseArea

saw-jan: yes true, same issue happens for settings and other topbar items.

saw-jan: Except `Add Account` button item. Other topbar items have `Accessible.role: Accessible.PageTab` which is the cause for the issue and we need `Accessible.onPressAction` to for those tabs as well.

TheOneRing: Aaaah I see, thank you.

saw-jan:
I will do the follow up PR for other tabs

src/gui/qml/AccountBar.qml
saw-jan: Follow up PR https://github.com/opencloud-eu/desktop/pull/1045

Issue #861, opened 2026-04-07
[QA] GUI automation tests in appium


saw-jan:
Porting squish gui tests into appium with behave.

POC: https://github.com/opencloud-eu/desktop/pull/857

Docs:
- behave: https://behave.readthedocs.io/en/stable/gherkin/
- linux driver: https://community.kde.org/Selenium#Writing_Tests

### Things to consider:
- restructure features, steps and helpers
- move step defs to `test/gui/steps`
- convert step defs and data tables into behave format
- replace squish APIs with appium or custom ones
- setup hooks and configs
- identify smoke tests (add `@smoke` tag) and fix them

#### Replace squish API with appium
Appium APIs: https://appium.github.io/appium.io/docs/en/about-appium/api
Selector format:
```py
from appium.webdriver.common.appiumby import AppiumBy as By

class LoginWizard:
  SOME_ELEMENT = SimpleNamespace(by=By.NAME, selector="Login")
```

### 💡 Guidelines for Porting GUI Tests
1. Move feature file from `tst_***` into its own suite inside `features/` dir
2. Remove `@skip` tag from the smoke test scenarios
3. Fix step definition and table if necessary (⚠️ behave has different [step param](https://behave.readthedocs.io/en/latest/tutorial/#step-parameters) and table parsing)
4. Replace necessary element selector. Preference (In order):
      - `BY.NAME`
      - `BY.ACCESSIBILITY_ID`
      - `BY.XPATH`
5. Replace Squish API with appium
6. Make the test scenario pass
7. Make a PR targeting `test/port-to-appium` branch

### Tasks
- [x] Identify smoke test scenarios (#867)
- [x] Enable scenarios (smoke scenarios):
  - [x] tst_activity (@prashant-gurung899)
    - #868
  - [x] tst_addAccount (@saw-jan)
    - #857
    - #869
    - #876
  - [x] tst_deleteFilesFolders (@anon-pradip)
    - #872
  - [x] tst_editFiles (@anon-pradip)
    - #875
  - [x] tst_loginLogout @anon-pradip  https://github.com/opencloud-eu/desktop/pull/877
  - [x] tst_moveFilesFolders @prashant-gurung899 
     - https://github.com/opencloud-eu/desktop/pull/884
  - [x] tst_removeAccountConnection (@prashant-gurung899)
    - #878
  - [x] tst_spaces (@saw-jan)
    - #883 
  - [x] tst_syncing @prashant-gurung899 
[...]

PR #884, opened 2026-04-27
test(gui): port moveFilesFolders test suite to appium

prashant-gurung899:

Enable following scenarios:

```feature
Scenario: Rename a file and a folder
```

```feature
Scenario: Move files from one folder to another
```

### DIFF
--- a/test/gui/steps/server_context.py
+++ b/test/gui/steps/server_context.py
@@ -1,6 +1,5 @@
 from behave import given as Given, then as Then
 from sure import ensure
-import parse
 from helpers.api import provisioning, webdav_helper as webdav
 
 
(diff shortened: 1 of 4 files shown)


saw-jan (approved):

test/gui/steps/server_context.py
```diff
@@ -24,14 +24,12 @@ def step(context, user_name, resource_type, resource_name):
...
-#         text_content,
-#         content,
-#         f"File '{file_name}' should have content '{content}' but found '{text_content}'",
-#     )
+@Then('as "{user_name}" the file "{file_name}" should have the content "{content}" in the server')
+def step(context, user_name, file_name, content):
+    text_content = webdav.get_file_content(user_name, file_name)
+    expect(text_content).to.equal(
+        content
+    )
```
anon-pradip: If we need custom message for assertion we can use like this:
```
with ensure('{0}  should have content "{1}" but found "{2}"', file_name, content, text_content):
  text_content.should.equal(content)
```

anon-pradip (approved):

PR #889, opened 2026-04-28
test(gui): enable syncing tests

prashant-gurung899:

Enabled the following syncing tests:

```feature
Scenario: Sync only one folder from the server
```

```feature
Scenario: Unselected subfolders are excluded from local sync
```

Added a new Python library called [PyAutoGui](https://pyautogui.readthedocs.io/en/latest/index.html) for GUI automation(mouse, keyboard, screenshots, etc.)

### DIFF
--- a/test/gui/features/sync-resources/syncResources.feature
+++ b/test/gui/features/sync-resources/syncResources.feature
@@ -82,7 +82,7 @@ Feature: Syncing files
         But the folder "simple-folder" should not exist on the file system
         And the folder "large-folder" should not exist on the file system
 
-    @skipOnWindows @smoke @skip
+    @skipOnWindows @smoke
     Scenario: Sync only one folder from the server
         Given user "Alice" has created folder "simple-folder" in the server
         And user "Alice" has created folder "large-folder" in the server
--- a/test/gui/features/sync-resources/syncResources.feature
+++ b/test/gui/features/sync-resources/syncResources.feature
@@ -571,7 +571,7 @@ Feature: Syncing files
         And as "Brian" file "Shares/simple-folder/simple.pdf" should exist in the server
         And as "Brian" the file "Shares/simple-folder/uploaded-lorem.txt" should have the content "overwrite openCloud test text file" in the server
 
-    @skipOnWindows @smoke @skip
+    @skipOnWindows @smoke
     Scenario: Unselected subfolders are excluded from local sync
         Given user "Alice" has created folder "test-folder" in the server
         And user "Alice" has created folder "test-folder/sub-folder1" in the server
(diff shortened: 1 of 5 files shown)


anon-pradip:
My findings:

## Issue

When testing the selective sync folder selection in the OpenCloud Desktop client, Appium could not toggle the checkboxes in the Qt tree widget. The test would fail with:

```
selenium.common.exceptions.WebDriverException: Message: element's actions list didn't contain SetFocus. The element may be malformed
```

## Root Cause

The Qt `QTreeWidget` checkbox is implemented using `QTreeWidgetItem::setCheckState()`. While the checkbox is visually rendered, the **AT-SPI accessibility bridge does not expose the toggle action** to automation frameworks.

Investigation showed:
- `checkable: true` - element is checkable
- `actions: None` - no actions exposed via AT-SPI
- `role: None` - role not exposed
- Standard `click()` selects the row but doesn't toggle the checkbox
- `send_keys(" ")` fails because the element lacks proper AT-SPI action interface

## What We Tried

| Method | Result |
|--------|--------|
| `element.click()` | Selects row, doesn't toggle checkbox |
| `TouchAction` / `app().tap()` | Not supported in this environment |
| `send_keys(" ")` / `send_keys(Keys.SPACE)` | Element doesn't support keyboard actions |
| `execute_script()` (JavaScript) | Command not supported |
| XPath to find checkbox child | No separate checkbox element exists |
| `pyatspi`, `pyautogui`, `Xlib` | Not available in test environment |

## Solution

Use **xdotool** - a command-line X11 automation tool that interacts directly with the X window system, bypassing AT-SPI limitations.

```python
# After Appium click fails, use xdotool:
subprocess.run(['xdotool', 'mousemove', str(x), str(y)], check=True)
subprocess.run(['xdotool', 'click', '1'], check=True)
```

Where `(x, y)` is the checkbox position (10-15px from left edge of the row).

## Implementation

The fix was added to `SyncConnectionWizard.select_or_unselect_folders_to_sync()`:

1. Try Appium click (selects row)
2. If checkbox not toggled, use xdotool to click at checkbox position

test/gui/features/sync-resources/syncResources.feature
```diff
@@ -580,6 +580,7 @@ Feature: Syncing files
         When the user unselects the following folders to sync in "Choose what to sync" window:
             | folder                  |
             | test-folder/sub-folder2 |
+        And the user waits for the files to sync
```
prashant-gurung899, line 583: Had to add this because it would fail on the following step as the synchronization took a little more time:
```feature
But the folder "test-folder/sub-folder2" should not exist on the file system
```

prashant-gurung899:
> My findings:

@anon-pradip 
Thank you for giving an alternative solution for checking/unchecking on checkboxes.
But for now, we will go on with a python library called `PyAutoGui`.

saw-jan (approved):

Issue #619, opened 2025-10-24
Linux: Flatpak package dristribution


hd-px:
Currently, the only installation option in your documentation for Linux is via AppImage [0].
I would therefore like to request whether there are plans to distribute a Flatpak package as well.
I think a lot of customers would appreciate this, as Flatpak, alongside Snap, is now considered the standard and is directly supported by the major package management tools [1] [2].

[0] https://docs.opencloud.eu/docs/user/desktop-client/installation/#linux-installation-appimage
[1] https://apps.kde.org/discover/
[2] https://gitlab.gnome.org/GNOME/gnome-software

Issue #517, opened 2025-09-03
Feature Request: Flatpak Support


N-Hertstein:
While it is amazing that linux is supported with the desktop app, updates still leave something to be desired.

My recommendation is creating a flatpak. It works across most linux distributions, supports updates trough the flatpak package manager and that's about it.

No need to create a .deb .rpm and pacman file.
It's abot as universal as an appimage (afaik)

NeoTheFox:
https://github.com/NeoTheFox/eu.opencloud.desktop

Here is a WIP manifest. It builds, but as @JannoTjarks mentioned, I would love to see an official build, since unverified apps are very hard to get through the security audits.

TheOneRing:
I know that you all love flatpack. But stating that appimages are deprecated technology and don't run on modern systems is just not true.
Our appimages use the latest runtime which does not require libfuse2.

All you need to do is to make the appimage executable and run it.
And yes you don't get a fancy package manager and have to update it manually.


We don't plan to provide any flatpack support in the near future.

Issue #211, opened 2025-04-16
Release for arch64?


handtrixx:
Hello guys,
I love your project. I am working on a MacBook with Asahi Linux so for the client I would need a arm64/arch64 release, which currently is not available. I was able to compile it on my own (https://github.com/handtrixx/opencloud-eu-desktop-arm64/releases/tag/v1.0.0-rc.3_arch64) but would prefer not to have to do that in future.
Do you plan to release an official arch64 client for linux? 
That would help raspi owners and so on as well.

TheOneRing:
It's not (yet) on the official roadmap, but on my personal.

handtrixx:
Dear @TheOneRing thank you very much for the quick response, and the at least partially good news. Ok, then I will try to automate the process a bit more to have something until there is an official release.

sallyhaj:
Thank you so much for your support @handtrixx .
I am only using Arm64 PC with Linux.

handtrixx:
Hi @sallyhaj , I agree with you. Maybe instead of AppImages Flatpaks would be better, since the handling of updates is easier with them. To figure out what kind of error you ran into, would take me some time as well and I am currently writing another tool. But, maybe good news: If you were already able to get the output as you described, there should be a folder "bin" beside the .rpm file. in that folder you find the executable "opencloud". Normally you should be able to run this, independent of your distribution. To get later updates you would to go through the same procedure again. Does that work for you?

sallyhaj:
Thank you so much @handtrixx for all your amazing efforts. Do you think that, Fathub will make the verification icon to the app? And is your efforts will be considered as an official app as AppImage?

sallyhaj:
@handtrixx Are you planning to update the repo to latest release?

handtrixx:
@sallyhaj : Just got the info message that its outdated today first time :-) yes i still have a mac on linux which is arch64, so its worth a look. give me 1-2 days. i will reply here

handtrixx:
@sallyhaj : pardon, took a bit long since i wanted this to be my first project executed by my local AI instance. But here you go, tested for x86_64 and aarch64 (your case): https://github.com/handtrixx/opencloud-eu-desktop-arm64/releases/tag/latest . let me know if there are any issues I didn't see.

sallyhaj:
It is functioning as normal. Thank you so much @handtrixx I home they will accept to add aarch64 as official support.

PR #892, opened 2026-05-01
Feat: Flatpak support

bryce-hoehn:
Addresses https://github.com/opencloud-eu/desktop/issues/211, https://github.com/opencloud-eu/desktop/issues/517, https://github.com/opencloud-eu/desktop/issues/619, and partially addresses https://github.com/opencloud-eu/desktop/issues/244.

Based on the manifest created by [NeoTheFox](https://github.com/NeoTheFox/eu.opencloud.desktop), with an additional Github Actions workflow for building the flatpak. Currently the build just uploads as a zip into the workflow artifacts. It shouldn't be too much extra work to add a step for publishing to Flathub, but it would need to come from the official team.


bryce-hoehn:
It should now be ready for review. I've added github actions to automatically update & publish the flatpak packages alongside releases. If I understand correctly, all that needs done now is creating the initial PR to Flathub, which needs to be done by the OpenCloud team. Documentation: https://docs.flathub.org/docs/for-app-authors/submission#submission-pr

In compliance with Flathub's [application ID requirements](https://docs.flathub.org/docs/for-app-authors/requirements#application-id), I also went ahead and replaced eu.opencloud-eu.desktop with eu.opencloud-eu.OpenCloudDesktop throughout the application. This might be a breaking change for existing installs, but could be mitigated by selectively applying the name only on flatpak builds.

Let me know if there is anything you'd like me to change!

bryce-hoehn:
I've noticed a few minor things that might need to be addressed before approval.

1. Currently the workflow updates the manifest after a release is triggered. This technically means the source code bundle in each release would still have the manifest from the previous release. Not sure if this really makes a difference for Flathub.
2. If you wanted to commit to changing the app id throughout all platforms, we could probably implement a migration script to move existing directories with the old app id.

bryce-hoehn:
Just pushed a fix to the workflow. Now the flatpack manifest should update automatically whenever VERSION.cmake is updated, while the publish to flathub job still only runs on release.

eu.opencloud.OpenCloudDesktop.metainfo.xml
```diff
@@ -0,0 +1,41 @@
...
+  <summary>Sync and collaborate on your desktop or laptop</summary>
+  <metadata_license>CC0-1.0</metadata_license>
+  <developer id="eu.opencloud">
+    <name>Heinlein Group</name>
+  </developer>
+  <description>
+    <p>The OpenCloud desktop client keeps photos and documents always up to date, enabling you to work like you always did.
+Any file you add, modify or delete in the synced folders on your desktop or laptop will show up, change or disappear on the server and all other connected devices.</p>
+  </description>
+  <icon type="stock">eu.opencloud.OpenCloudDesktop</icon>
```
bkmgit:
```suggestion
  <icon>eu.opencloud.OpenCloudDesktop</icon>
```
When using appstream-util validate-relax --nonet from [appstream glib](https://github.com/hughsie/appstream-glib) get an error
```
• tag-invalid           : stock icon is not valid [eu.opencloud.OpenCloudDesktop]
```

javsanpae:
Hi! Is this effort still ongoing? I'd really appreciate having a Flatpak package for OpenCloud. Let me know if there's anything I can help with!

TheOneRing:
We currently don't plan to officially support Flatpak. Therefore, this pull request can't be merged.
You have to understand, we are supporting three different platforms already with a small team. All platforms use exact the same version of Qt etc.

However, we'd like to support the community, if possible, to provide Flatpak builds. For example, with the recent integration of AppStream metadata.

Peethiplatsch:
Being a noob, but why did you choose AppImage over Flatpak which is also supported on multiple Linux Distributions and is properly updateable etc. AppImage for me is really like a portable EXE under Windows with all its pros and cons. For a tool like OpenCloud the pros do not apply. But I am not a software dev nor someone who creates a package. So I don't know what the pros and cons are from your end and I am really interested in it since there are always two sites of a story :)

TheOneRing:
Exactly, it's as easy as a Windows installer or a Mac Bundle, you just stuff files into an archive.
No sandbox, not portals, super easy.
And as a sync application, we need to file system access anyhow....

PR #905, opened 2026-05-14
ci: fix woodpecker warnings

saw-jan:
Fixes the following warnings:
Screenshot from 2026-05-14 09-30-32

### DIFF
--- a/.woodpecker/notification.yaml
+++ b/.woodpecker/notification.yaml
@@ -7,12 +7,19 @@ variables:
     from_secret: oc_ci_url
 
 when:
-  - event: [push, manual]
+  - event:
+      - push
+      - manual
     branch: ${CI_REPO_DEFAULT_BRANCH}
+    status:
+      - success
+      - failure
   - event: pull_request
+    status:
+      - failure
   - event: cron
     cron: nightly*
-  - status:
+    status:
       - success
       - failure
 
--- a/test/gui/environment.py
+++ b/test/gui/environment.py
@@ -44,14 +44,14 @@ def before_feature(context, feature):
 
 
 def after_step(context, step):
-    if step.status in [Status.failed, Status.error]:
-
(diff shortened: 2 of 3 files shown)


test/gui/environment.py
```diff
@@ -44,14 +44,14 @@ def before_feature(context, feature):
 
 
 def after_step(context, step):
-    if step.status in [Status.failed, Status.error]:
-
-        step_name = re.sub(r'[^a-zA-Z0-9_]', '_', step.name)
+    if step.status in [Status.failed, Status.error] and os.getenv("CI"):
+        scenario = context.scenario.name.lower()
```
saw-jan, line 48: use scenario name as file name instead of step name

.woodpecker/notification.yaml
```diff
@@ -7,12 +7,19 @@ variables:
...
+  - event:
+      - push
+      - manual
     branch: ${CI_REPO_DEFAULT_BRANCH}
+    status:
+      - success
+      - failure
   - event: pull_request
+    status:
+      - failure
```
prashant-gurung899, line 19: Do we have to repeat the status for every event now?

test/gui/environment.py
```diff
@@ -44,14 +44,14 @@ def before_feature(context, feature):
 
 
 def after_step(context, step):
-    if step.status in [Status.failed, Status.error]:
-
-        step_name = re.sub(r'[^a-zA-Z0-9_]', '_', step.name)
+    if step.status in [Status.failed, Status.error] and os.getenv("CI"):
```
prashant-gurung899, line 47: So the reports are only for CI now?

saw-jan: yeah, we can enable it by setting CI env while running the tests locally

.woodpecker/notification.yaml
saw-jan: looks like so.

prashant-gurung899 (approved):

PR #906, opened 2026-05-14
test(gui): capture and store video for failed scenarios

prashant-gurung899:

This PR adds a screenrecorder helper.
Adds a `before_scenario` hook to start recording and only save the video if the scenario fails.
"

Use the python packages: `mss` and `inputio-ffmpeg`

### DIFF
--- a/test/gui/environment.py
+++ b/test/gui/environment.py
@@ -5,6 +5,7 @@ import pyautogui
 from behave.model_core import Status
 from datetime import datetime
 
+from helpers import ScreenRecorder
 from helpers.ConfigHelper import init_config
 from helpers.api.provisioning import delete_created_users
 from helpers.SpaceHelper import delete_project_spaces
@@ -43,6 +44,11 @@ def before_feature(context, feature):
     init_config()
 
 
+def before_scenario(context, scenario):
+    if os.getenv("CI"):
+        ScreenRecorder.start_recording(scenario)
+
+
 def after_step(context, step):
     if step.status in [Status.failed, Status.error] and os.getenv("CI"):
         scenario = context.scenario.name.lower()
@@ -56,6 +62,11 @@ def after_step(context, step):
 
 
 def after_scenario(context, scenario):
+
+    # stop screen recording
+    if os.getenv("CI"):
+        ScreenRecorder.stop_recording(passed=scenario.status == Status.passed)
+
     # clean up sync dir
     if os.path.exists(get_config("clientRootSyncPath")):
         for entry in os.scandir(get_config("clientRootSyncPath")):
--- a/test/gui/helpers/ScreenRecorder.py
+++ b/test/gui/helpers/ScreenRecorder.py
@@ -1,0 +1,82 @@
+import os
+import re
+import threading
+import time
+import mss
+import numpy as np
+import imageio_ffmpeg
+from datetime import datetime
+
+from helpers.ConfigHelper import get_config
+
+
+_recording_thread = None
+_stop_event = threading.Event()
+_video_path = None
+
+
+def _build_video_path(scenario):
+    safe_name = re.sub(r"[^a-zA-Z0-9_]", "_", scenario.name)
+    timestamp = datetime.now().strftime("%d-%b-%Y_%H-%M-%S")
+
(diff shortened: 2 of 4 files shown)


test/gui/environment.py
```diff
@@ -43,8 +47,13 @@ def before_feature(context, feature):
     init_config()
 
 
+def before_scenario(context, scenario):
+    if os.getenv("CI") == "true":
```
saw-jan: just check the existence of CI env
```suggestion
    if os.getenv("CI"):
```

test/gui/environment.py
```diff
@@ -43,8 +47,13 @@ def before_feature(context, feature):
...
+        screenRecorder.start_recording(scenario)
+
+
 def after_step(context, step):
-    if step.status in [Status.failed, Status.error] and os.getenv("CI"):
+    if step.status in [Status.failed, Status.error] and os.getenv("CI") == "true":
```
saw-jan:
```suggestion
    if step.status in [Status.failed, Status.error] and os.getenv("CI"):
```

test/gui/environment.py
```diff
@@ -56,6 +65,11 @@ def after_step(context, step):
 
 
 def after_scenario(context, scenario):
+
+    # stop screen recording
+    if os.getenv("CI") == "true":
```
saw-jan:
```suggestion
    if os.getenv("CI"):
```

test/gui/environment.py
prashant-gurung899: That way, any value to the variable CI is going to be true, isn't it?

saw-jan: yeah, if CI env is present then we consider it as CI environemt.
the actual value of `CI` is `woodpecker`.

prashant-gurung899: Ahhhh okay. Makes sense now. :+1:

test/gui/helpers/ScreenRecorder.py
```diff
@@ -0,0 +1,82 @@
+import os
+import re
+import threading
+import time
+from datetime import datetime
+
+import mss
+import numpy as np
+import imageio_ffmpeg
+from helpers.ConfigHelper import get_config
```
saw-jan, line 10:
```suggestion
import time
import mss
import numpy as np
import imageio_ffmpeg
from datetime import datetime

from helpers.ConfigHelper import get_config
```

test/gui/helpers/ScreenRecorder.py
saw-jan, line 1: `screenRecorder.py` -> `ScreenRecorder.py`

saw-jan (commented):
✨ I really need this for debugging. ❤️

test/gui/helpers/ScreenRecorder.py
```diff
@@ -0,0 +1,82 @@
...
+_recording_thread = None
+_stop_event = threading.Event()
+_video_path = None
+
+
+def _build_video_path(scenario):
+    safe_name = re.sub(r"[^a-zA-Z0-9_]", "_", scenario.name)
+    timestamp = datetime.now().strftime("%d-%b-%Y_%H-%M-%S")
+
+    recordings_dir = os.path.join(get_config("guiTestReportDir"), "recordings")
```
saw-jan, line 22: Need to show video reports links in CI. Similar to what we have for screenshots:
https://github.com/opencloud-eu/desktop/blob/2c583ca9a7f69bc26160f76882259d06d229d53e/test/gui/woodpecker/gui_test_reports.sh#L12-L22

test/gui/woodpecker/gui_test_reports.sh
```diff
@@ -20,3 +20,15 @@ if [[ -n "$screenshots" ]]; then
 else
   echo "No screenshots found."
 fi
+
+recordings=$(mc find s3/$REPORT_PATH/recordings/ 2>/dev/null || true)
+if [[ -n "$recordings" ]]; then
+  echo "Recordings:"
```
saw-jan, line 27: Need line break
```suggestion
  echo ""
  echo "Recordings:"
```

saw-jan (approved):
👍 recording is working ✨

PR #923, opened 2026-05-28
Pass the socket path to openVFS using the -s switch

dragotin:
Requires https://github.com/opencloud-eu/openvfs/pull/42/ in openVFS to read the socket path


src/plugins/vfs/openvfs/vfs_openvfs.cpp
```diff
@@ -210,14 +210,15 @@ void OpenVFS::startImpl(const VfsSetupParams &params)
     connect(_openVfsProcess, &QProcess::finished, this, [logPrefix, this] { qCFatal(lcOpenVFS) << logPrefix() << "finished" << _openVfsProcess->exitCode(); });
     connect(_openVfsProcess, &QProcess::started, this, [logPrefix, this] {
         qCInfo(lcOpenVFS) << logPrefix() << u"started";
-        // TODO:
-        // give it time to mount
+
```
TheOneRing: We still need a more elegant method than just wait for a random amount of time, so the todo is still valid in a way.

TheOneRing (changes requested):

PR #926, opened 2026-06-02
test(ci): add @skip tag to exclude scenario from CI execution

Asmitapaudel:
This change adds the @Skip tag to specific test scenarios to exclude them from CI execution.

Previously, non-smoke tests also required @smoke tag to run in CI, which caused confusion. This change separates CI inclusion (@smoke) and exclusion (@skip) for clearer test control.

---
Now the workflow will be :
- skip all the tests that are not tagged @smoke
- fix and enable skipped tests
- in CI, run tests that are not marked @skip

@saw-jan @prashant-gurung899  Should I keep the @smoke tag now or not?


saw-jan:
>@saw-jan @prashant-gurung899 Should I keep the @smoke tag now or not?

`@smoke` tags should be preserved. They indicate smoke test scenarios

anon-pradip (approved):
Now when migrating, we can remove `skip` tag and have that test run in CI 🚀

PR #929, opened 2026-06-02
Remove useless "undefined" in UI

### LINKED ISSUE #950: Spaces without a quota limit display "undefined" as the storage line
## Summary
Ran into this when packaging OpenCloud Desktop for Fedora COPR: 

For a Space whose quota is unlimited (server reports `total == 0`), the folder/Space list shows the literal text **`undefined`** where the storage usage line should be usage only, or be hidden. The account otherwise connects and syncs normally; only the storage label is wrong.

## Steps to reproduce

1. Connect the client to an OpenCloud server whose Spaces have **no quota limit** (capabilities: `"spaces":{"max_quota":0,…}`; `GET /graph/v1.0/me/drives` returns drives whose `quota.total` is `0`/absent).
2. Open the account view that lists the Spaces.

## Expected behavior
- no storage line (or an "unlimited" indication) for unlimited Spaces.

## Actual behavior
- the storage line reads `undefined`.

## Setup
- OpenCloud Desktop 3.0.3 (commit `72fe8f`), virtual files plugin: off
- Fedora 44 w/ Qt 6.11.1, OpenSSL 3.5.7, Wayland, Linux x86_64
- Server: OpenCloud 4.0.7, Spaces enabled, `max_quota: 0`
- Packages with patch in https://copr.fedorainfracloud.org/coprs/greve/opencloud-desktop/

## Root cause

`src/gui/folderstatusmodel.cpp`, `FolderStatusModel::data()`, `Roles::Quota`:

```cpp
case Roles::Quota: {
    qint64 used{};
    qint64 total{};
    if (auto *space = f->space()) {
        const auto quota = space->drive().getQuota();
        if (quota.isValid()) {
            used = quota.getUsed();
            total = quota.getTotal();
        }
    }
    if (total <= 0) {
        return {};   // <-- invalid QVariant()
    }
    return tr("%1 of %2 used").arg(Utility::octetsToString(used), Utility::octetsToString(total));
}
```

The role is consumed in `src/gui/qml/FolderDelegate.qml` as a typed **string**:

```qml
required property string quota
...
text: folderDelegate.quota
visible: folderDelegate.quota && !folderDelegate.overallText
```

When `total <= 0`, `data()` returns an **invalid `QVariant()`**. Assigned to a `required property string`, that becomes the JS value `undefined`, which stringifies to the literal `"undefined"`. The `visible:` guard then sees a non-empty (truthy) string and **shows** the label. An empty `QString` would instead coerce to `""`, which the guard treats as false and hides as a minimal fix.

The invalid-variant → `string` coercion is stricter on newer Qt (seen with Qt 6.11); the underlying defect: returning an invalid variant for a role bound to a typed `string` property is version-independent.

## Suggested fix (one line)

```diff
         if (total <= 0) {
-            return {};
+            return QString();
         }
```

This hides the storage line for unlimited Spaces via the existing `visible:` guard. (A follow-up could instead show "X used" or an "Unlimited" label, but the minimal correctness fix is to stop rendering `undefined`.)

guruz:
Screenshot 2026-06-02 at 14 38 11

Hide the "undefined" ^

### DIFF
--- a/src/gui/folderstatusmodel.cpp
+++ b/src/gui/folderstatusmodel.cpp
@@ -298,7 +298,7 @@ QVariant FolderStatusModel::data(const QModelIndex &index, int role) const
             }
         }
         if (total <= 0) {
-            return {};
+            return QString();
         }
         return tr("%1 of %2 used").arg(Utility::octetsToString(used), Utility::octetsToString(total));
     }


src/gui/folderstatusmodel.cpp
```diff
@@ -298,7 +298,7 @@ QVariant FolderStatusModel::data(const QModelIndex &index, int role) const
             }
         }
         if (total <= 0) {
-            return {};
+            return QStringLiteral("");
```
TheOneRing:
```suggestion
            return QString();
```

TheOneRing (changes requested):
^

src/gui/folderstatusmodel.cpp
guruz: > In Qt, the main difference between creating a QString and using QStringLiteral is runtime performance. QStringLiteral generates the string at compile-time and avoids the expensive heap allocation and character conversion that occurs when a standard QString constructor is called dynamically at runtime.

is this not true anymore? :)

guruz: There is an @ogoffart blog post ... https://woboq.com/blog/qstringliteral.html

TheOneRing: Yes, but not for empty strings.
In our code QString() is correct as it creates a null string.
(It can actually be more complicated https://github.com/KDE/clazy/blob/master/docs/checks/README-empty-qstringliteral.md )

guruz: ugggh ... thanks

guruz: I tested it, it doesnt print the "undefined"then so good

TheOneRing (approved):

PR #944, opened 2026-06-11
refactor(gui-test): fix config parsing and use the report helper

saw-jan:
- Fix config parsing and normalization
- Use ReportHelper for report paths
- Screen recording is disabled by default
  - use `RECORD_VIDEO_ON_FAILURE=true` to enable the screen recording
  - enabled in CI
  - NOTE: the limit of screen recordings is 5. This is to prevent large test reports size

### DIFF
--- a/test/gui/environment.py
+++ b/test/gui/environment.py
@@ -25,11 +25,7 @@ def before_feature(context, feature):
 
 
 def before_scenario(context, scenario):
-    if (
-        os.getenv("CI")
-        and get_config("record_video_on_failure")
-        and not hit_screenrecord_limit()
-    ):
+    if get_config("record_video_on_failure") and not hit_screenrecord_limit():
         ScreenRecorder.start_recording(normalize_scenario_title(scenario.name))
     elif hit_screenrecord_limit():
         print("[INFO] Screen recording limit reached.")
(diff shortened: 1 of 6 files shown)


saw-jan:
^ @Asmitapaudel

test/gui/environment.py
```diff
@@ -13,61 +10,50 @@
...
 def before_feature(context, feature):
     init_config()
 
 
 def before_scenario(context, scenario):
-    if os.getenv("CI"):
-        ScreenRecorder.start_recording(scenario)
+    if (
+        os.getenv("CI")
+        and get_config("record_video_on_failure")
```
prashant-gurung899: Do we need both of the envs?

prashant-gurung899:
LGTM :+1:

test/gui/environment.py
saw-jan: hmm, may be not.

test/gui/environment.py
```diff
@@ -13,61 +10,49 @@
...
-
 def before_feature(context, feature):
     init_config()
 
 
 def before_scenario(context, scenario):
-    if os.getenv("CI"):
-        ScreenRecorder.start_recording(scenario)
+    if (
+        and get_config("record_video_on_failure")
```
prashant-gurung899:
```suggestion
        get_config("record_video_on_failure")
```

prashant-gurung899 (approved):

saw-jan:
I will deal with flaky tests separately.

PR #949, opened 2026-06-16
CI: Pin the Visual Studio release

TheOneRing:


saw-jan:
@TheOneRing should we port this to `stable-3.0` branch as well?
Also, build in stable-3.0 branch doesn't work, so I tried to do this https://github.com/opencloud-eu/desktop/pull/951 (don't know if that's the proper fix)

TheOneRing:
Currently, there is no plan for another 3.x release, so I would refrain from a backport unless needed.

saw-jan:
Okay, thank you.
I have necessary test changes ready to be ported in a PR for release-3.0. I will continue on that when the stable-3.0 becomes active again.

PR #951, opened 2026-06-17
build: unpin craft-core

saw-jan:


saw-jan:
not need for now.
Ref: https://github.com/opencloud-eu/desktop/pull/949#issuecomment-4729493931

PR #955, opened 2026-06-21
fix(creds): keep the job queue alive across transient token refreshes

### LINKED ISSUE #900: Per-drive sync stays in Aborted after a single 401 on /me/drives, even when subsequent requests succeed
## Steps to reproduce

1. Sign in to a desktop client and let one or more drives sync to "Up to date".
2. Restart the IdP while the client is running (e.g. a Keycloak pod restart, or a database failover that cascades into Keycloak). The restart is brief; on our cluster the rollout takes about 35 seconds.
3. One of the periodic `/graph/v1.0/me/drives` polls returns 401 because the session was mid-refresh when the IdP was unavailable.
4. Wait for the IdP to recover. Subsequent `/me/drives` polls return 200 (verified server-side, see log evidence below).

## Expected behavior

- Per-drive sync resumes automatically once 200s come back, optionally with a short "retrying" indicator while the drive re-discovers.

## Actual behavior

- The Personal drive (and any other drive polled during the 401 window) shows **"Aborted due to: This space is currently unavailable"** indefinitely.
- The user has to manually **Reconnect** (gear menu) or **Log out / Log in** to clear the wedge. Files are not lost since they live on the server, but sync does not resume on its own.
- Roughly 95% of subsequent calls succeed server-side, the client just does not re-attempt the affected drive.

## Setup

- **Operating system**: Linux Mint (separate user reports indicate the same on macOS 3.0.x)
- **Client version**: OpenCloud Desktop 3.0.3.2073 (AppImage on Linux)
- **Server**: OpenCloud with a custom storage driver. The bug is not storage-driver-specific, it is an OIDC-discovery state-machine issue, so any deployment that ever restarts its IdP during an active session should reproduce it.
- **Log**: anonymised server-side log evidence is below. A client-side log can be provided on request.

## Server-side log evidence (anonymised)

```
07:35:54  GET /graph/v1.0/me/drives -> 401 (token expired during IdP restart)
07:35:55  GET /graph/v1.0/me/drives -> 200 (1 second later, same client)
07:36:24  GET /graph/v1.0/me/drives -> 200
07:37:24  GET /graph/v1.0/me/drives -> 200
... continues 200 for hours ...
```

Throughout this window the client UI continues to show the Personal drive as **Aborted**. The 401 is correct, spec-compliant behaviour from the IdP. The bug is that the desktop client treats it as a terminal state for the affected drive instead of retrying discovery on a backoff.

## Suggested fix

When a drive is marked Aborted because of an auth failure on `/me/drives`, re-attempt discovery on a backoff schedule (for example 30 s, 1 min, 2 min, 5 min, 10 min) for as long as the **account itself** is still authenticated. Surface a "retrying" state in the UI rather than a terminal failure. If the account itself is de-authenticated, the existing "Log in" prompt is the correct fallback.

This is architecturally adjacent to, but distinct from, the proactive token-refresh request in #763 / opencloud-eu/opencloud#919. Proactively refreshing before expiry would prevent the 401 from happening in the first place. This issue is about graceful recovery once a 401 *does* happen, which it will, for example whenever the IdP restarts during an active session.

## Related observations

Validated 2026-05-10 against v3.0.3.2073 on Linux Mint. All three live in the per-folder / per-drive status pipeline and were surfaced by the same Keycloak-restart exercise. They may share a root cause, or at least share a triage owner.

### 1. No `Reconnect` action in the gear menu on Linux

The gear menu next to the account only offers **Log out**. Operator runbooks and community guides that say "click Reconnect" assume a macOS/Windows-only UI. Either Linux is missing the action or the cross-platform documentation is wrong.

### 2. Slow auto-recovery from "Queued"

Concrete repro:

1. Sign in, drop a 100 KB file into a synced folder, wait for the green check.
2. Trigger an IdP rollout while a *new* file is being uploaded ("0 seconds left, file 0 of 1" or actively transferring).
3. The client transitions to **Queued** with the loading animation during the rollout (about 35 s in our measurement).
4. IdP returns 200 on `/me/drives` within roughly 1 second of being back up. Verified server-side.
5. The client stays in **Queued** for more than 10 minutes despite no further 4xx/5xx from the server. Only **Log out / Log in** clears it (Reconnect is not available on Linux, see observation 1).

This appears to be the same symptom as the now-closed #661 ("Sync gets stuck in 'Queued' (v3.0.0-beta.5)"). #661 was auto-closed by the stale-bot without a fix, and the bug still reproduces on 3.0.3.2073.

### 3. "0 seconds left, file 0 of 1" progress widget never dismisses

Every small-file upload reproduces it. Drop a file into a synced folder, the file uploads successfully (verified in the web UI), and the per-folder progress widget continues to show "0 seconds left, file 0 of 1" indefinitely. It survives until a state-refresh event (IdP restart, Log out / Log in). Probably unrelated to the Aborted bug above, but worth bundling for the same release given the shared per-folder status pipeline.

## Related issues

- #661, "Sync gets stuck in 'Queued' (v3.0.0-beta.5)", closed `not planned` by the stale-bot on 2026-02-18. Same symptom as observation 2 above.
- #763, "Feature: Refresh access token before it times out", closed `not planned` by the stale-bot on 2026-02-04. Architecturally adjacent: proactive refresh would prevent the 401 that triggers the Aborted state.
- opencloud-eu/opencloud#919, server-side counterpart of #763.

### LINKED ISSUE #948: JobQueue::clear() aborts all in-flight sync jobs (uploads & downloads) when OAuth .well-known is rate-limited
When syncing — **in either direction** — a transient rate-limit response (`{"error":"Too many requests"}`) from the auth server's `.well-known/openid-configuration` endpoint causes `JobQueue::clear()` to fire, aborting **all** pending jobs regardless of type: `GETFileJob` (downloads), `SimpleNetworkJob` / `POST` (uploads), and TUS transfers. Running jobs are critically deleted. The sync engine declares the run complete, so aborted files are never retried and the client shows no error. Quitting and restarting repeatedly eventually works through the backlog, but it takes many cycles.

Reproduced on **two different Macs** (MBP M1 Max and Mac Mini M4), both running client **3.0.3.2073** against OpenCloud 7.1.0. On the MBP, four consecutive log segments (spanning ~3 minutes of restarts) each show the identical failure: upload jobs aborted by the same rate-limit cascade. The Mini captured the download-side failure cleanly.

## Steps to Reproduce

1. Create a space with several hundred files across nested directories
2. Connect a macOS desktop client (3.0.3.2073) to sync that space, HTTP logging enabled
3. Observe discovery + reconciliation complete normally, jobs queued
4. A `.well-known/openid-configuration` fetch returns `{"error":"Too many requests"}`
5. `JobQueue::clear()` aborts all pending jobs — `POST` (uploads), `GET` (downloads), and running jobs
6. Client shows no error; directories exist but files are missing; subsequent syncs skip already-"synced" items

## MBP Upload Failures (4 consecutive restarts, each aborted)

```
11:53:51  JobQueue::clear → Abort SimpleNetworkJob ×21 (POST uploads) + 12 critical deletions
11:54:33  JobQueue::clear → Abort SimpleNetworkJob ×21 (POST uploads) + 10 critical deletions
11:55:05  JobQueue::clear → Abort SimpleNetworkJob ×22 (POST uploads)
11:55:32  JobQueue::clear → Abort SimpleNetworkJob ×21 (POST uploads) + 3 critical deletions
```

## Mini Download Failure (initial sync)

```
11:59:08  .well-known → {"error":"Too many requests"} → JobQueue::clear
          Abort GETFileJob ×20 + Deleting running job ×5
          Result: 88/100s of files synced, directories empty
```

## Environment

- macOS 26.4.1 (Mac Mini) / 25.5.0 (MBP), client 3.0.3.2073
- OpenCloud 7.1.0 behind Caddy, HTTP/2 enabled
- Pocket ID (OIDC) auth at separate auth server

## Workaround

Quitting and restarting the client repeatedly eventually works through the sync backlog, as each cycle gets a few more files through before hitting the rate limit again.

## Related

- opencloud-eu/desktop#900 (auth error cascading into permanent wedge)
- opencloud-eu/desktop#734 (server dislikes client behavior / HTTP 420 under load)
- opencloud-eu/desktop#812 (desktop client freezes during large sync — may share root cause)

bernardgut:
## Problem

The desktop client can **silently drop files** when an OAuth access token is refreshed mid-sync. Afterwards directories exist on the server but files inside are missing, and later syncs *skip* the already-recorded items — silent, persistent loss. This is the mechanism behind #900 ("Aborted after a single 401 on `/me/drives`") and #948 (a transient `.well-known/openid-configuration` response aborts all jobs).

## Root cause

`HttpCredentials::refreshAccessTokenInternal()` splits refresh errors into *transient* (schedule a retry) and *terminal* (give up after `TokenRefreshMaxRetries`). The **transient** branch schedules the retry but then *also* `Q_EMIT authenticationFailed();`. `Account` connects that signal to `JobQueue::clear()` (`src/libsync/account.cpp`), which `abort()`s every queued upload. Aborted-but-unsent jobs are `deleteLater()`d without ever calling `done()`, so no error/retry flag is set; `SyncEngine::finalize()` then commits the journal as "All Finished" and the dropped files are never re-discovered.

So a *transient* refresh — which the client intends to retry — is treated as a *terminal* failure **for the job queue**.

## Fix

Emit `authenticationFailed()` only on the terminal branch. On a transient error the job queue stays blocked (it was blocked by `authenticationStarted()`) across the scheduled retry and resumes via `fetched() -> JobQueueGuard::unblock()` on success. The early-`return` is restructured into an explicit `if (terminal) / else (transient)` so the two outcomes are clear.

## Why it's safe

`authenticationFailed()` has two consumers — `account.cpp` (`JobQueue::clear()`) and `cmd.cpp` (`qFatal()` in `opencloudcmd`). Both should react only to a *terminal* failure; the CLI now waits out the bounded retry instead of dying on a transient blip. The retry is bounded by `TokenRefreshMaxRetries == 3`, and the terminal branch still emits the signal, so the queue cannot stay blocked forever.

## Reproduction

OpenCloud + per-drive OAuth (Keycloak), `mirall/3.0.3.2073`, sync a folder large enough that the run outlives one access-token lifetime. At the expiry boundary the proxy logs a single `401` on `/me/drives`; the client goes "Aborted", directories are created but files are missing, and re-syncing does not recover them.

## Testing

I don't have a Qt build environment to run the suite here, and there is currently **no test that drives `HttpCredentials::refreshAccessToken`** (`test/testoauth` covers `AccountBasedOAuth`; `test/testjobqueue` covers the queue primitives). The right regression test would deliver a *transient* `refreshError` and assert (a) `authenticationFailed()` is **not** emitted and the queue stays blocked, then (b) after `TokenRefreshMaxRetries` it **is** emitted. Glad to add it if you can point me at the preferred seam (extend the `testoauth` `FakeAM`/`FakeErrorReply` harness to instantiate `HttpCredentials`, or an integration test in `testsyncengine`).

Fixes #900
Refs #948


bernardgut:
## Regression test added (f44ea97)

Added two regression tests to `test/testoauth/testoauth.cpp` covering the fix's observable contract:

**`testTransientRefreshDoesNotEmitAuthFailed()`** — verifies that a transient `refreshError` (e.g. `TimeoutError`) does *not* emit `authenticationFailed()`, so the job queue stays blocked (not cleared) and in-flight uploads are preserved.

**`testTerminalRefreshEmitsAuthFailed()`** — verifies that after `TokenRefreshMaxRetries` (3) consecutive incrementing errors (`ContentNotFoundError`, timeout=0s), `authenticationFailed()` *is* emitted along with `fetched()`, triggering the terminal logout path.

### Prove-the-test-bites result

| Scenario | testTransient | testTerminal |
|----------|:---:|:---:|
| Bug reintroduced (`Q_EMIT authenticationFailed()` in transient branch) | **FAIL** | **FAIL** |
| Fix applied (current) | **PASS** | **PASS** |

Full suite: 31/31 pass, 0 regressions.

Built and tested in `opencloudeu/desktop-client-build:ubuntu-24.04-qt6.10` with `-DBUILD_TESTING=ON`.

TheOneRing (approved):
Thanks for your contribution, if I dissect the git blame correctly that error exists for at least 3 years.

PR #960, opened 2026-07-01
fix: harden recoverable-error handling so the sync self-heals (#898)

### LINKED ISSUE #898: TUS uploads can get permanently stuck after failed resume; stale uploadinfo is not reset since 3.0
Since 3.0, large Windows batch uploads (~20GB, many files) become unreliable/stuck. Before 3.0 this worked.

Suspected regression:
- #333 enabled HTTP/2 by default.
- Account starts with _http2Supported=true.
- Folder::loadSyncOptions() returns 20 network jobs for HTTP/2.
- OwncloudPropagator::maximumActiveTransferJob() now uses std::max(3, ceil(parallelJobs/2)), which allows 10 active transfer jobs with HTTP/2 instead of capping at 3.
- Backend advertises TUS creation-with-upload with default 10MB max chunk size.
- #637 removed repeated upload error reset logic for UploadInfo/TUS.

Repro:
1. Windows client 3.0.
2. OpenCloud backend with default TUS upload protocol.
3. Drop ~20GB/many files into sync folder.
4. Observe upload timeouts/stuck/retries/incomplete uploads.
5. Retest with OPENCLOUD_HTTP2_DISABLED=1 or with transfer concurrency capped at 3.

Expected:
Large batch upload eventually completes.

Actual:
Uploads become unreliable/stuck after 3.0.

@guruz @dragotin

bernardgut:
## Problem

Beyond the OAuth token-refresh wedge (#955), a couple of recoverable error conditions break the desktop client's self-healing during a large sync — a sync should always converge to complete on the next pass / re-login, but these bugs either abort the whole run or wedge a file. They bite hardest on long, multi-day syncs where transient errors are near-certain.

### 1. A *transient* network error aborts the entire sync run — not just the file
`classifyError()` (`src/libsync/owncloudpropagator_p.h`) maps **every** `QNetworkReply` error in `(NoError, UnknownProxyError]` to `FatalError`, which propagates up to `propagator()->abort()` and aborts the **whole run**. On a long sync a single timeout / connection-reset / temporary-failure on one file kills the entire pass, and the expensive discovery has to restart from scratch.

Now the recoverable connectivity errors (`Timeout`, `ConnectionRefused`, `HostNotFound`, `TemporaryNetworkFailure`, `NetworkSessionFailed`, proxy refused/closed/timeout) are a per-file `NormalError` that sets `anotherSyncNeeded`, so only that file is retried and the run continues. Genuinely fatal cases (TLS handshake, proxy auth, redirect loops, …) keep `FatalError`.

### 2. TUS resume wedges on a 409 Upload-Offset mismatch (#898)
A stale/diverged TUS resume gets HTTP 409. It used to fall through to `commonErrorHandling` and the upload wedged (the file hangs near 100%, never completes; recovery needs a manual remove + re-add of the folder). `PropagateUploadFileTUS::slotChunkFinished()` now routes a 409 through the **same** HEAD-to-get-current-offset recovery path it already uses for timeouts, resuming from the server's canonical `Upload-Offset`. `classifyError()` also classifies 409 as a recoverable `SoftError` + `anotherSyncNeeded` as a fallback.

## Tests

Two new tests, both **bug-bites-verified** (each fails against the pre-fix code, passes after):

- `test/testclassifyerror.cpp` (`ClassifyError`) — unit-pins the `classifyError` contract: transient connectivity errors → `NormalError` + retry (not `FatalError`); `409` → `SoftError` + retry; genuinely-fatal errors (e.g. TLS handshake) stay `FatalError`. Pre-fix: `Actual FatalError ≠ Expected NormalError`, `Actual NormalError ≠ Expected SoftError`.
- `test/testselfheal.cpp` (`SelfHeal`) — FakeFolder integration test proving the self-heal property end-to-end: a transient network error injected on one file must not abort the whole run; the healthy files queued after it still upload. Pre-fix this test fails (the run aborts on the first file, the healthy files queued behind it never sync).

Full client `ctest` is green (**33/33**) on the CI build image (`opencloudeu/desktop-client-build:ubuntu-24.04-qt6.10`).

Fixes #898

### DIFF
--- a/src/libsync/owncloudpropagator_p.h
+++ b/src/libsync/owncloudpropagator_p.h
@@ -47,11 +47,40 @@ inline SyncFileItem::Status classifyError(
     }
 
     if (nerror > QNetworkReply::NoError && nerror <= QNetworkReply::UnknownProxyError) {
-        // network error or proxy error -> fatal
-        return SyncFileItem::FatalError;
+        // A *transient* connectivity error on a single file must NOT abort the whole sync
+        // run. The blanket FatalError below returns up to propagator()->abort(), which wedges
+        // a large multi-day sync on one network blip. Treat the recoverable connectivity
+        // errors as a per-file NormalError and request another pass so the file is
+        // re-discovered and retried; keep FatalError only for genuinely fatal cases
+        // (TLS handshake, proxy auth, redirect loops, ...).
+        switch (nerror) {
+        case QNetworkReply::ConnectionRefusedError:
+        case QNetworkReply::HostNotFoundError:
+        case QNetworkReply::TimeoutError:
+        case QNetworkReply::TemporaryNetworkFailureError:
+        case QNetworkReply::NetworkSessionFailedError:
+        case QNetworkReply::ProxyConnectionRefusedError:
+        case QNetworkReply::ProxyConnectionClosedError:
+        case QNetworkReply::ProxyTimeoutError:
+            if (anotherSyncNeeded != nullptr) {
+                *anotherSyncNeeded = true;
+            }
+            return SyncFileItem::NormalError;
+        default:
+            // network error or proxy error -> fatal
+            return SyncFileItem::FatalError;
+        }
     }
 
     switch (httpCode) {
+    case 409:
+        // "Conflict" -- e.g. a TUS Upload-Offset mismatch on resume (opencloud-eu/desktop#898).
+        // Recoverable: the TUS path resumes from the server's offset; here we ensure the file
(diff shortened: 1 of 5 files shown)


TheOneRing:
I know it's pesky work, but could you please split that in two merge requests.

bernardgut:
Done, split in two as requested:

- #970: classifyError treats transient network errors as per-file retries instead of aborting the whole run
- #971: TUS resumes from the server's offset on a 409 (fixes #898)

Both are rebased on current main, independent of each other, and each has its own regression tests. Closing this one. Thanks for the review!

Issue #553, opened 2025-09-24
[QA] Implement code analyzer tool to catch unused code


anon-pradip:
Sometimes there are code snippets (methods or classes or any code) that are used nowhere but still we are not aware of that unused code. For enhancing the detection of such unused code, it will be far more easier if we implement and use any code analyzer tool. I am listing here some of the tools available in the market:
- [vulture](https://github.com/jendrikseipp/vulture)
- [pyflakes](https://pypi.org/project/pyflakes/) - finds unused imports and unused local variables (in addition to many other programmatic errors).
- [coverage](https://pypi.org/project/coverage/) - finds unused code more reliably than Vulture, but requires all branches of the code to actually be run.
- [uncalled](https://pypi.org/project/uncalled/) - finds dead code by using the abstract syntax tree (like Vulture), regular expressions, or both.
- [dead](https://pypi.org/project/dead/) - finds dead code by using the abstract syntax tree (like Vulture).

CC ^ @TheOneRing @individual-it

individual-it:
At the same time we could add linter and code-style tools. @Salipa-Gurung should have good ideas what tools to use

Salipa-Gurung:
The combination of **Ruff** and **Vulture** provides a fast, modern, low-maintenance baseline covering linting, formatting, and dead code detection. **Pylint** may be introduced selectively where deeper architectural or complexity analysis is required.

1. **Ruff**
    pros: Very fast, replaces multiple linters, strong default rules, unified linting and formatting
    cons: Limited deep architecture/code-smell analysis 
3. **Vulture**
    pros: detects globally unused (dead) code, lightweight, fast (uses static code analysis)
    cons: can have false positives, focused only on dead code 
5. **Pylint**
    pros: Strong code-smell and complexity detection, useful design insights
    cons: Slower, can be noisy without tuning, overlaps with Ruff

CC: @saw-jan @individual-it

saw-jan:
Implemented ruff and vulture

PR #968, opened 2026-07-08
chore(gui-test): add ruff formatter

Asmitapaudel:
part of https://github.com/opencloud-eu/desktop/issues/553

## Summary

* Add Ruff formatter to the project.
* Apply Ruff's default formatting while preserving existing string quotes (quote-style = "preserve"). 

### Ruff Formatter
Purpose: Automatically formats Python code to a consistent style.
It changes how the code looks, not how it behaves.

It fixes things like:
 *  Indentation
  * Spacing
  * Blank lines
  * Line wrapping
  * Quote style (if configured)

## Formatter commands

```bash
# Format files in place
uv run ruff format .
```

```bash
# Check whether files are already formatted
uv run ruff format . --check
```

```bash
# Show formatting differences without modifying files
uv run ruff format . --diff
```

### DIFF
--- a/.woodpecker/python-lint.yaml
+++ b/.woodpecker/python-lint.yaml
@@ -1,0 +1,22 @@
+variables:
+  - &astral_uv_image 'astral/uv:python3.12-trixie'
+when:
+  - branch:
+      - main
+      - stable-*
+    event:
+      - push
+      - manual
+  - event: pull_request
+
+workspace:
+  base: /woodpecker
+  path: desktop
+
+steps:
+  - name: lint-test-code
+    image: *astral_uv_image
+    commands:
+      - cd test/gui
+      - uv sync --only-group lint 
+      - make python-lint
--- a/test/gui/Makefile
+++ b/test/gui/Makefile
@@ -34,7 +34,7 @@ check-uv:
 .PHONY: uv-install
 uv-install: check-uv
 	uv venv --system-site-packages --allow-existing
-	uv sync --frozen
+	uv sync --frozen --all-groups
 
 .PHONY: install-chromium
 install-chromium: check-uv
@@ -45,14 +45,12 @@ run-atpsi-webdriver: check-uv
 	bash woodpecker/run_atspi_webdriver.sh
 
 .PHONY: python-lint
-python-lint:
-	black --check --diff .
-	python3 -m pylint --rcfile ./.pylintrc $(PYTHON_LINT_PATHS)
+python-lint: check-uv
+	uv run --no-sync ruff format . --check
 
 .PHONY: python-lint-fix
-python-lint-fix:
-	black .
-	python3 -m pylint --rcfile ./.pylintrc $(PYTHON_LINT_PATHS)
+python-lint-fix: check-uv
+	uv run --no-sync ruff format .
 
 .PHONY: gherkin-lint
 gherkin-lint:
--- a/test/gui/pyproject.toml
+++ b/test/gui/pyproject.toml
@@ -7,7 +7,6 @@ dependencies = [
     "appium-python-client==5.3.*",
     "behave==1.3.*",
     "behave-html-pretty-formatter==1.16.*",
-    "black==24.3.*",
     "flask==3.0.* ; sys_platform == 'linux'",
     "imageio-ffmpeg==0.6.*",
     "mss==9.*",
@@ -27,3 +26,11 @@ dependencies = [
     "requests==2.32.*",
     "sure==2.0.*",
 ]
+
+[dependency-groups]
+lint = [
+    "ruff>=0.15.20",
+]
+
+[tool.ruff.format]
+quote-style = "preserve"
--- a/test/gui/helpers/AppHelper.py
+++ b/test/gui/helpers/AppHelper.py
@@ -95,7 +95,7 @@ def create_app_session():
         f'{get_config("app_path")} -s {command_args} --logdebug',
     )
     options.set_capability('appium:environ', get_app_env())
-    options.set_capability('timeouts', { 'implicit': get_config('min_timeout') * 1000 })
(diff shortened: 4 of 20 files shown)


test/gui/pyproject.toml
```diff
@@ -26,4 +26,11 @@ dependencies = [
     "pywin32==305 ; sys_platform == 'win32'",
     "requests==2.32.*",
     "sure==2.0.*",
-]
\ No newline at end of file
+]
+
+[dependency-groups]
+dev = [
+    "ruff>=0.15.20",
```
saw-jan, line 32: lint would be a fit group name for the linters
```suggestion
[dependency-groups]
lint = [
    "ruff>=0.15.20",
```

test/gui/pyproject.toml
saw-jan, line 1: we need a command on how to run the linter.

test/gui/pyproject.toml
```diff
@@ -26,4 +26,12 @@ dependencies = [
     "pywin32==305 ; sys_platform == 'win32'",
     "requests==2.32.*",
     "sure==2.0.*",
-]
\ No newline at end of file
+]
+
+[dependency-groups]
+lint = [
```
saw-jan, line 31: we should have a command to install this as well.
maybe here:
https://github.com/opencloud-eu/desktop/blob/c23507bb0d6faea6e5eb72cacf7497ce0140ec1b/test/gui/Makefile#L7-L9

test/gui/pyproject.toml
saw-jan: we need a command in the Makefile

saw-jan, line 1: Also, add a CI workflow to check the formatting

test/gui/Makefile
```diff
@@ -44,15 +44,13 @@ install-chromium: check-uv
 run-atpsi-webdriver: check-uv
 	bash woodpecker/run_atspi_webdriver.sh
 
-.PHONY: python-lint
-python-lint:
-	black --check --diff .
-	python3 -m pylint --rcfile ./.pylintrc $(PYTHON_LINT_PATHS)
+.PHONY: ruff-formatter
+ruff-formatter:
```
saw-jan: we can keep the old name
```suggestion
.PHONY: python-lint
python-lint:
```

test/gui/Makefile
```diff
@@ -34,7 +34,7 @@ check-uv:
 .PHONY: uv-install
 uv-install: check-uv
 	uv venv --system-site-packages --allow-existing
-	uv sync --frozen
+	uv sync --frozen --group lint
```
saw-jan: hmm, will this install all the dependencies?

.woodpecker/build.yaml
saw-jan, line 1: revert the file changes

.woodpecker/code-analysis.yaml
```diff
@@ -0,0 +1,26 @@
+variables:
+  - &build_image 'opencloudeu/desktop-client-build:ubuntu-24.04-qt6.10'
```
saw-jan: do we need to use this? maybe we can use any container with uv

.woodpecker/code-analysis.yaml
saw-jan, line 1: I would rename the file: `code-analysis.yaml` -> `python-lint.yaml`

.woodpecker/code-analysis.yaml
```diff
@@ -0,0 +1,26 @@
...
+      
+depends_on:
+  - cache-python
+
+workspace:
+  base: /woodpecker
+  path: desktop
+
+steps:
+  - name: ruff-format
```
saw-jan:
```suggestion
  - name: lint-test-code
```

test/gui/Makefile
Asmitapaudel: yes, it installs all dependencies as well as everything under the lint group.

saw-jan: now I wonder, why have separate lint dependency group then :thinking: ?

.woodpecker/python-lint.yaml
```diff
@@ -0,0 +1,25 @@
+variables:
+  - &uv_astral_image 'astral/uv:python3.12-trixie'
```
saw-jan:
```suggestion
  - &astral_uv_image 'astral/uv:python3.12-trixie'
```

.woodpecker/python-lint.yaml
```diff
@@ -0,0 +1,25 @@
...
+  - branch:
+      - main
+      - stable-*
+    event:
+      - push
+      - manual
+  - event: pull_request
+      
+depends_on:
+  - cache-python
```
saw-jan: it should not depend on this workflow if not using the cache

test/gui/Makefile
```diff
@@ -45,14 +45,12 @@ run-atpsi-webdriver: check-uv
 	bash woodpecker/run_atspi_webdriver.sh
 
 .PHONY: python-lint
-python-lint:
-	black --check --diff .
-	python3 -m pylint --rcfile ./.pylintrc $(PYTHON_LINT_PATHS)
+python-lint: check-uv
+	uv run --no-project ruff format . --check
```
saw-jan: do we need `--no-project`?
```suggestion
	uv run ruff format . --check
```

Asmitapaudel: Yes, `uv run` also works like `uv sync`. If we just run `uv run`, it installs the other dependencies as well. So don't install the other dependencies; just use the existing environment with the `lint` group only.

It throws an error if we don't use --no-project:
https://ci.opencloud.rocks/repos/11/pipeline/395/27#L51

saw-jan: maybe we should `--no-sync` because we are disabling sync action

saw-jan: same should be for lint-fix command

saw-jan (approved):
👍 ✨

PR #969, opened 2026-07-08
Fix account stuck in "Connecting" state after network switch

paolostivanin:
When switching networks (e.g. WiFi → mobile hotspot), a running `ConnectionValidator`'s HTTP request could hang on the dead socket. The guard in `checkConnectivity()` would then block all subsequent reconnection attempts, leaving the account permanently stuck in "Connecting".

Two fixes:
- Abort any stale `ConnectionValidator` when reachability or captive-portal state changes, so a fresh validation can start immediately, and clear the access manager's connection cache so new requests open fresh sockets on the new interface.
- Add a 60-second hard timeout to `ConnectionValidator` to abort hung requests as a safety net.

Split out of #858 so it can be reviewed independently.

Tested on openSUSE Tumbleweed x86_64.

### DIFF
--- a/src/gui/accountstate.cpp
+++ b/src/gui/accountstate.cpp
@@ -90,6 +90,18 @@ AccountState::AccountState(AccountPtr account)
         case NetworkInformation::Reachability::Site:
             [[fallthrough]];
         case NetworkInformation::Reachability::Unknown:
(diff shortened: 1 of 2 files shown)


TheOneRing (approved):
Interesting find, thank you for your contribution!

PR #970, opened 2026-07-10
fix(libsync): treat transient network errors as per-file retries, not whole-run aborts

bernardgut:
Splitting #960 in two as requested. This is the first half.

## Problem

classifyError() treats every QNetworkReply error up to UnknownProxyError as FatalError, which aborts the whole sync run via propagator()->abort(). So a single timeout or connection reset on one file kills the entire pass and discovery starts over. On long syncs this happens all the time, and the sync never converges.

## Fix

Recoverable connectivity errors (timeout, connection refused, host not found, temporary network failure, network session failed, proxy refused/closed/timeout) now return a per-file NormalError and set anotherSyncNeeded. Only that file is retried and the run continues. Genuinely fatal cases (TLS handshake, proxy auth, redirect loops) keep FatalError.

## Tests

Both fail against the pre-fix classifyError and pass with it:

- test/testclassifyerror.cpp pins the classification contract.
- test/testselfheal.cpp is a FakeFolder integration test: a transient error injected on one file must not abort the run, and the healthy files behind it still upload.

Full ctest on the CI image: 32/33. The one failure is testfolderwatcher, which fails on my build host for environment reasons (inotify instance limit exhausted, unrelated to this change).

Together with #955 this makes long syncs converge instead of wedging. None of these fixes are in a release yet. If you would consider a 3.0.4, I am happy to prepare the stable-3.0 backport PRs.


TheOneRing (approved):
Thanks, sounds very reasonable

PR #971, opened 2026-07-10
fix(libsync): resume a TUS upload from the server's offset on a 409 mismatch

### LINKED ISSUE #898: TUS uploads can get permanently stuck after failed resume; stale uploadinfo is not reset since 3.0

bernardgut:
Second half of the #960 split. Fixes #898.

## Problem

A stale TUS resume gets HTTP 409 (Upload-Offset mismatch) and falls into commonErrorHandling, where the upload wedges: the file hangs near 100% and only a manual remove and re-add of the folder recovers it (#898).

## Fix

- PropagateUploadFileTUS::slotChunkFinished() now handles a 409 like a timeout: HEAD the upload to get the server's current offset and resume from there.
- As a fallback, classifyError() maps 409 to SoftError and sets anotherSyncNeeded, so the file is retried instead of being finalized with a silent gap.

## Tests

test/testtusresume.cpp pins the classifyError fallback (fails pre-fix, passes with the fix). The transport recovery reuses the existing timeout path rather than adding a new one.

Full ctest on the CI image: 31/32. The one failure is testfolderwatcher, environment-only on my build host (inotify limit, unrelated).

Fixes #898


TheOneRing (approved):

TheOneRing:
Thank you for your contribution

TheOneRing:
The crashing unit test is unrelated

PR #978, opened 2026-07-16
fix: decode JWT segments as Base64url (RFC 7515)

### LINKED ISSUE #977: OIDC login fails with "The audience of the id_token did not contain …" when the id_token payload's base64url encoding contains `-` or `_`
## Steps to reproduce

Deterministic, no IdP needed:

```cpp
// A valid base64url-encoded JWT payload: {"aud":"OpenCloudDesktop","pic":"https://x.example/aa?v=1"}
// The '?' at zero-based payload offset 53 (≡ 2 mod 3) encodes to sextet 63 → '_' in base64url.
QByteArray part = "eyJhdWQiOiJPcGVuQ2xvdWREZXNrdG9wIiwicGljIjoiaHR0cHM6Ly94LmV4YW1wbGUvYWE_dj0xIn0";
qDebug() << QByteArray::fromBase64(part);                                // JWT::JWT() does this: garbage after ".../aa"
qDebug() << QByteArray::fromBase64(part, QByteArray::Base64UrlEncoding); // correct JSON
```

`QJsonDocument::fromJson` on the first decode fails, so `JWT::payload()` is an empty object.

Real-world path (how I hit it):

1. OpenCloud with an external IdP (PocketID 2.11.0 here) and a public OIDC client with ID `OpenCloudDesktop`.
2. Add the account in the desktop client, complete the browser login. The IdP includes a claim whose base64url encoding contains `-` or `_` (in my case a `gravatar_url` claim whose `?v=…` query string lands at the right byte alignment).
3. Token exchange succeeds (IdP returns 200 with valid tokens), but the client shows: **"Login Error — The audience of the id_token did not contain \"OpenCloudDesktop\""**, although the id_token's actual audience is exactly `["OpenCloudDesktop"]` (verified by performing the same PKCE exchange manually and decoding with a spec-compliant decoder).

## Expected behavior
- JWT segments are decoded as Base64url (RFC 7515 §2) and login succeeds.
- If an id_token genuinely cannot be parsed, the error says so, rather than reporting an audience mismatch.

## Actual behavior
- `JWT::JWT()` decodes segments with `QByteArray::fromBase64(part)` standard-alphabet Base64. Qt's default `IgnoreBase64DecodingErrors` mode skips `-`/`_` as invalid characters ([Qt docs](https://doc.qt.io/qt-6/qbytearray.html#fromBase64)), dropping one 6-bit sextet each; all following bytes are regrouped and decode to garbage. The payload JSON no longer parses, `IdToken::aud()` returns `{""}`, and the check in `oauth.cpp` reports the audience error.

## Analysis

- Decoder: [`src/libsync/creds/jwt.cpp` line 23 @ v3.0.3](https://github.com/opencloud-eu/desktop/blob/v3.0.3/src/libsync/creds/jwt.cpp#L23); unchanged on [main @ c23507b](https://github.com/opencloud-eu/desktop/blob/c23507bb0d6faea6e5eb72cacf7497ce0140ec1b/src/libsync/creds/jwt.cpp#L23). Introduced in e3f0a0f0 ("Verify login of correct user by reading the id_token").
- Error site: [`src/libsync/creds/oauth.cpp` line 373 @ v3.0.3](https://github.com/opencloud-eu/desktop/blob/v3.0.3/src/libsync/creds/oauth.cpp#L373).
- Note `JWT::serialize()` in the same file already encodes with `Base64UrlEncoding | OmitTrailingEquals` and only the decode direction uses the wrong alphabet.
- Believe only the **payload** segment matters for this failure: the signature is stored, not decoded, and the header isn't consulted on this path.
- Why I think this isn't failing for everyone: sextets 62/63 (`-`/`_`) are rare in base64url of plain-ASCII JSON. For ASCII payloads they mainly arise from `?` or `>` at payload offset ≡ 2 (mod 3), or from non-ASCII UTF-8 bytes. Compact tokens usually decode fine by luck; tokens with URL-with-query-string claims (picture/avatar URLs), display names with non-ASCII characters, etc. hit it depending on alignment. That makes the failure look random/IdP-specific while actually being content-dependent.
- Verified with Qt 6.11.1 (behavior matches documented semantics of `QByteArray::Base64Option`, stable across Qt 6).

## Suggested fix

Decode as Base64url and surface parse failures instead of letting them masquerade as audience mismatches:

```cpp
auto parse = [](const QByteArray &part) {
    const auto result = QByteArray::fromBase64Encoding(part,
        QByteArray::Base64UrlEncoding | QByteArray::AbortOnBase64DecodingErrors);
    if (!result) {
        return QJsonObject{};
    }
    return QJsonDocument::fromJson(*result).object();
};
```

plus (ideally) a distinct "the id_token could not be parsed" error in `oauth.cpp` before the `aud()` check. With only the decoder fixed, a genuinely malformed token would still produce the misleading audience message. (Since the `JWT` is constructed as a temporary there, the simplest check is on the resulting `IdToken`'s payload being empty.)

One deliberate interoperability consequence worth noting: strict Base64url decoding will reject tokens from IdPs that (non-compliantly) emit standard-alphabet `+`/`/` in JWT segments. That matches RFC 7515, but it is a behavior change for such IdPs.

Possibly related: #217 reports the audience error with Authelia when roles are configured. Plausibly the same decode failure triggered by different payload content, though I haven't looked into it.

## Setup
- Windows 11 25H2 (OS build 26200.8875), OpenCloud desktop 3.0.3.2073 (mirall/3.0.3)
- Server: OpenCloud (Docker, external IdP: PocketID 2.11.0)
- Log: IdP-side log shows `POST /api/oidc/token` → 200 with valid tokens; client then reports the audience error.

willyp713:
Fixes #977

## What

`JWT::JWT()` decoded token segments with `QByteArray::fromBase64(part)`, which uses the standard Base64 alphabet. JWT segments are Base64url (RFC 7515 section 2). Qt's default `IgnoreBase64DecodingErrors` mode silently skips `-` and `_` as invalid characters and corrupts every byte after them. The payload then fails JSON parsing, `IdToken::aud()` returns an empty string, and OIDC login fails with a misleading "The audience of the id_token did not contain …" error. See #977 for the full analysis, including why the failure depends on token content and looked IdP-specific.

## Changes

Commit 1:

- `jwt.cpp`: decode segments with `Base64UrlEncoding | AbortOnBase64DecodingErrors` via `fromBase64Encoding`. A header or payload segment that is not valid base64url now yields an empty object, and so an invalid JWT, instead of silently corrupted claims. `serialize()` already used `Base64UrlEncoding`; only parsing was wrong.
- `test/testjwt.cpp` (new): regression tests. They cover a payload whose base64url encoding contains both `-` and `_`, a serialize/parse round-trip, structurally malformed tokens, and strict rejection of an invalid character inside an otherwise valid segment (the permissive decode wrongly accepted that case). Against the unpatched code, three of the four fail; with this change all pass.

Commit 2:

- `oauth.cpp`: an unparseable id_token previously fell through to the audience check and produced the misleading audience error. The token exchange handler now retains the parsed `JWT` and reports "The id_token could not be parsed" before checking the audience.
- `test/testoauth/testoauth.cpp`: new `testUnparseableIdToken` corrupts the header and payload segments in turn, and expects the parse error rather than a login or the audience message.

## Notes

- Strict Base64url decoding rejects tokens from IdPs that emit standard-alphabet `+` or `/` in JWT segments. RFC 7515 forbids those characters, but this is a behavior change for such non-compliant IdPs.
- Tested on Linux with Qt 6.8.2. I compiled the `testjwt` target standalone and ran it against both the patched and unpatched decoder; I did not run the full suite build in my environment.

### DIFF
--- a/src/libsync/creds/jwt.cpp
+++ b/src/libsync/creds/jwt.cpp
@@ -13,6 +13,9 @@
  */
 #include "jwt.h"
 
+#include <QLoggingCategory>
+
+Q_LOGGING_CATEGORY(lcJwt, "sync.credentials.jwt", QtInfoMsg)
 
 OCC::JWT::JWT(const QByteArray &jwt)
 {
@@ -24,6 +27,7 @@ OCC::JWT::JWT(const QByteArray &jwt)
     auto parse = [](const QByteArray &part) {
         const auto decoded = QByteArray::fromBase64Encoding(part, QByteArray::Base64UrlEncoding | QByteArray::AbortOnBase64DecodingErrors);
         if (!decoded) {
+            qCWarning(lcJwt) << "Failed to Base64url-decode JWT segment, status:" << static_cast<int>(decoded.decodingStatus);
             return QJsonObject{};
         }
         return QJsonDocument::fromJson(*decoded).object();
--- a/src/libsync/creds/jwt.cpp
+++ b/src/libsync/creds/jwt.cpp
@@ -20,7 +20,14 @@ OCC::JWT::JWT(const QByteArray &jwt)
     if (parts.size() != 3) {
         return;
     }
-    auto parse = [](const QByteArray &part) { return QJsonDocument::fromJson(QByteArray::fromBase64(part)).object(); };
+    // JWT segments are Base64url encoded (RFC 7515 section 2), not standard Base64
+    auto parse = [](const QByteArray &part) {
+        const auto decoded = QByteArray::fromBase64Encoding(part, QByteArray::Base64UrlEncoding | QByteArray::AbortOnBase64DecodingErrors);
+        if (!decoded) {
+            return QJsonObject{};
+        }
+        return QJsonDocument::fromJson(*decoded).object();
+    };
     _header = parse(parts[0]);
     _payload = parse(parts[1]);
     _signauture = parts[2];
--- a/src/libsync/creds/jwt.h
+++ b/src/libsync/creds/jwt.h
@@ -15,9 +15,10 @@
 #include <QJsonDocument>
 #include <QJsonObject>
 #include <QString>
+#include "opencloudsynclib.h"
 
 namespace OCC {
-class JWT
+class OPENCLOUD_SYNC_EXPORT JWT
 {
 public:
     /*
--- a/src/libsync/creds/oauth.cpp
+++ b/src/libsync/creds/oauth.cpp
@@ -789,7 +789,7 @@ void AccountBasedOAuth::restore()
     auto idTokenJob = _account->credentialManager()->get(idTokenC());
     connect(idTokenJob, &CredentialJob::finished, this, [idTokenJob, this] {
         if (idTokenJob->error() == QKeychain::EntryNotFound) {
-            qCWarning(lcOauth) << u"idToken token token credential not found";
+            qCWarning(lcOauth) << u"idToken token credential not found";
         } else if (idTokenJob->error() != QKeychain::NoError) {
             Q_EMIT result(Error);
             return;
--- a/src/libsync/creds/oauth.cpp
+++ b/src/libsync/creds/oauth.cpp
@@ -341,7 +341,8 @@ void OAuth::startAuthentication()
                     const QString refreshToken = getRequiredField(data, QStringLiteral("refresh_token"), &fieldsError).toString();
                     const QString tokenType = getRequiredField(data, QStringLiteral("token_type"), &fieldsError).toString().toLower();
                     const QUrl messageUrl = QUrl::fromEncoded(data[QStringLiteral("message_url")].toByteArray());
-                    auto idToken = IdToken(JWT(getRequiredField(data, QStringLiteral("id_token"), &fieldsError).toByteArray()).payload());
+                    const JWT jwt(getRequiredField(data, QStringLiteral("id_token"), &fieldsError).toByteArray());
+                    auto idToken = IdToken(jwt.payload());
 
 
                     auto reportError = [socket, this](const QString &errorReason) {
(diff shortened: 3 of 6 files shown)


dragotin:
Thank you for the patch, looks like an interesting finding. 

However, I needed [pr_978.patch](https://github.com/user-attachments/files/30094350/pr_978.patch) to make it compile over here, please double check and update the PR accordingly.

willyp713:
Thanks for the patch; applied in full and force-pushed. The export macro and Q_SLOTS fixes are squashed into the first commit, the enum name fix into the second, and the two cleanups (duplicated word in the log message, redundant QtTest include) are a separate third commit.

The misses came from verifying only testjwt locally, on a Linux build with default symbol visibility and without QT_NO_KEYWORDS. I re-ran testjwt with QT_NO_KEYWORDS defined: 6/6 pass.

dragotin (approved):
Looks now good to me, lets see what @TheOneRing thinks.

src/libsync/creds/jwt.cpp
```diff
@@ -20,7 +20,14 @@ OCC::JWT::JWT(const QByteArray &jwt)
     if (parts.size() != 3) {
         return;
     }
-    auto parse = [](const QByteArray &part) { return QJsonDocument::fromJson(QByteArray::fromBase64(part)).object(); };
+    // JWT segments are Base64url encoded (RFC 7515 section 2), not standard Base64
+    auto parse = [](const QByteArray &part) {
+        const auto decoded = QByteArray::fromBase64Encoding(part, QByteArray::Base64UrlEncoding | QByteArray::AbortOnBase64DecodingErrors);
+        if (!decoded) {
+            return QJsonObject{};
```
TheOneRing, line 31: Please add a log line including the `QByteArray::Base64DecodingStatus`

TheOneRing (changes requested):

src/libsync/creds/jwt.cpp
willyp713: Added a `qCWarning` under a new `sync.credentials.jwt` category logging the `Base64DecodingStatus` on decode failure. Status only (no segment bytes) to keep token contents out of the logs. `testjwt` still 6/6.

TheOneRing (approved):
The crash in the unit test is unrelated.
Thank you for your contribution.

PR #979, opened 2026-07-16
test(gui): configurable timeout for element search

saw-jan:
Implement configurable timeout while using `find_element`. Default element search timeout is `5s`. Now we can configure it on the go.

Usage:
```py
driver.find_element('name', 'my button', 10)  # 10 seconds
```

### DIFF
--- a/.woodpecker/build.yaml
+++ b/.woodpecker/build.yaml
@@ -55,6 +55,7 @@ steps:
 
   - name: build-client-for-ui-tests
     image: *build_image
+    pull: true # remove this once the nightlies run with the new build image
     commands:
       - mkdir -p build
       - cd build
--- a/.woodpecker/cache-python.yaml
+++ b/.woodpecker/cache-python.yaml
@@ -58,6 +58,7 @@ steps:
 
   - name: install-python-modules
     image: *build_image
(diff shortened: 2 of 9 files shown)


saw-jan:
@Asmitapaudel ^

anon-pradip (approved):
Just a minor thought, using `request` and `requests` in same file made me confused a bit at first sight. Can we just use  tweak `send_request` to `def send_request(url, method, body=None, headers=None, user=None, password=None, json=None):`
Then we can remove both `import requests` `import helpers.api.http_helper as request` and use like
```python
import helpers.api.http_helper as http_helper
...
def set_implicit_wait(timeout):
    session_id = app().session_id
    url = f'{get_config("webdriver_url")}/session/{session_id}/timeouts/implicit_wait'
    response = http_helper.send_request(url, 'POST', json={'ms': timeout * 1000})
    http_helper.assert_http_status(response, 200, 'Failed to set implicit timeout')
```

But this deserves a separate PR as other files will also be touched :thinking: ? Your call totally. Other things looks :ok_hand: to me

saw-jan:
this should be doable with minimal changes in this PR. :+1:

PR #980, opened 2026-07-17
Use environment variable to enable FUSE logging in openVFS

dragotin:
So far, the fuse log floods the log output of the client. Also name the logging category more speaking


src/plugins/vfs/openvfs/vfs_openvfs.cpp
```diff
@@ -216,9 +216,15 @@ void OpenVFS::startImpl(const VfsSetupParams &params)
...
     connect(_openVfsProcess, &QProcess::errorOccurred, this, [logPrefix, this] { qCWarning(lcOpenVFS) << logPrefix() << _openVfsProcess->errorString(); });
 
-    const QStringList pparams{u"-d"_s, u"-i"_s, openVFSConfigFilePath().toString(), u"-o"_s, xattrOwnerString(params.account->uuid()),
-                u"-s"_s, params.socketPath, params.root().toString()};
-    qCDebug(lcOpenVFS) << "Starting openvfs" << pparams;
+    QStringList pparams{u"-f"_s, u"-i"_s, openVFSConfigFilePath().toString(), u"-o"_s, xattrOwnerString(params.account->uuid()), u"-s"_s, params.socketPath,
+        params.root().toString()};
+
+    // Enable _lots_ of logging for fuse layer
+    if (qgetenv("OPENVFS_FUSE_LOG").length()) {
```
TheOneRing: I detest environment variables in ui applications.
Ideally, we could send the output to a logging category, but hooking the output of openvfs to a callback instead of directly dumping to the terminal caused a deadlock. Maybe we should investigate there.

```suggestion
    if (qEnvironmentVariableIntValue("OPENCLOUD_OPENVFS_LOG_FUSE") == 1) {
```

TheOneRing (changes requested):

src/plugins/vfs/openvfs/vfs_openvfs.cpp
dragotin: This is not about how we handle the logging, but to switch the debug output on in the fuse layer only if it is needed. That will be more performant than producing tons of log that nobody is interested in I assume.

dragotin: Should we rather use a config setting in the client config to enable/disable the fuse log?

TheOneRing: > This is not about how we handle the logging, but to switch the debug output on in the fuse layer only if it is needed. That will be more performant than producing tons of log that nobody is interested in I assume.

My point was that if we were able to properly redirect the input, we could introduce a logging category `sync.vfs.openvfs.fuselogs`.  Which would offer us a standardized switch to enable or disable those logs without yet another env var.

We currently have no concept of debug switches for plugins, so I don't think we should add it to the ui.

TheOneRing (approved):

Issue #217, opened 2025-04-18
Authelia: Problem with user roles


carlosjfcasero:
I have integrated Opencloud with Authelia. In web everything is working, but in desktop I get following error in server when I try to login:

```
opencloud  | 2025-04-18T18:50:48Z ERR Error mapping role names to role ids error="no roles in user claims" line=github.com/opencloud-eu/opencloud/services/proxy/pkg/userroles/oidcroles.go:84 request-id=58733e08-061f-4caa-876a-d68132cb6a09 service=proxy userid=e15359ea-ef03-4785-8cb3-564a93372e6d
opencloud  | 2025-04-18T18:50:48Z ERR Could not get user roles error="no roles in user claims" line=github.com/opencloud-eu/opencloud/services/proxy/pkg/middleware/account_resolver.go:174 service=proxy
```
It seems Opencloud is not retrieving correctly the claims from Authelia, but I don't know why. I have followed [integration with external OIDC doc](https://github.com/opencloud-eu/docs/blob/293dcb5b1c3869269ba14a2adb89efde5c2245ee/docs/admin/configuration/authentication-and-user-management/external-idp.md#integrating-external-openid-connect-identity-providers) and checked [troubleshooting page](https://github.com/opencloud-eu/docs/blob/293dcb5b1c3869269ba14a2adb89efde5c2245ee/docs/admin/configuration/authentication-and-user-management/keycloak.md#troubleshooting) but didn't find my issue.

My Authelia users configuration (only web and desktop part, used my ocis config as reference to build it) is
```
      - client_id: web
        client_name: 'OpenCloud web'
        public: true
        redirect_uris:
          - 'https://opencloud.tld/'
          - 'https://opencloud.tld/oidc-callback.html'
          - 'https://opencloud.tld/oidc-silent-redirect.html'
      - client_id: 'OpenCloudDesktop'
        client_name: 'openCloud desktop client'
        public: true
        scopes:
          - openid
          - groups
          - profile
          - email
          - offline_access
        redirect_uris:
          - http://127.0.0.1
          - http://localhost
        grant_types:
          - refresh_token
          - authorization_code
```

TheOneRing:
I think you might need to add `PROXY_ROLE_ASSIGNMENT_OIDC_CLAIM=groups` to the environment of your OpenCloud docker-compose.

carlosjfcasero:
I added mentioned line but nothing changed. I noticed this line that may help:
```
[...]

PR #847, opened 2026-03-17
Use webfinger for OIDC parameter discovery

kaivol:
Closes #811.

Client side implementation of the changes described in https://github.com/opencloud-eu/opencloud/blob/main/docs/adr/0003-oidc-client-config-discovery.md. 

Based on #776.

PR #983, opened 2026-07-20
fix: use a lowercase OAuth client identifier

roder:
## Summary

OpenCloud Desktop currently advertises a static OAuth client identifier
`OpenCloudDesktop`. Some identity providers normalize client identifiers
when registering or looking up an OAuth client, so the ID token's
`aud`, `azp`, and `client_id` claims end up containing the normalized
form (e.g. `openclouddesktop`).

`src/libsync/creds/oauth.cpp` performs an exact, case-sensitive
audience check:

    idToken.aud().contains(_clientId)

With the mixed-case configured ID and a normalized audience, the
desktop rejects its own valid token after a successful authorization
and token exchange. This change normalizes the static client
identifier to `openclouddesktop` so the request and the audience check
agree on a single canonical value.

The audience comparison itself is intentionally not changed. Per
RFC 6749 §2.2 and OpenID Connect Core 1.0 §2 / §3.1.3.7, client
identifiers are opaque strings and the comparison must remain exact.

## Background and prior discussion

This PR is the client-side follow-up to
[kanidm/kanidm#4473](https://github.com/kanidm/kanidm/issues/4473),
where the case-sensitivity contract between the IdP and the client was
discussed. The conclusion of that discussion — that the client must
use the normalized value consistently — is what this PR implements.

This PR may also be relevant to
[opencloud-eu/desktop#217](https://github.com/opencloud-eu/desktop/issues/217),
which reports a different symptom (`no roles in user claims` from the
server-side proxy) but involves a deployment that registers
`OpenCloudDesktop` with Authelia. The same exact-match audience check
on the desktop is plausibly a contributor there, but the original
issue does not diagnose that, and this PR is not presented as a
resolution of #217 — only as a candidate mitigation to be verified
by a maintainer or the original reporter.

## Changes

- `src/libsync/theme.cpp`: `Theme::oauthClientId()` returns
  `"openclouddesktop"` instead of `"OpenCloudDesktop"`.
- `test/testoauth/testoauth.cpp`: the synthetic ID token's `aud`,
  `azp`, and `client_id` claims are updated to match.

## Operator notes

Operators of OpenCloud Desktop deployments must register
`openclouddesktop` with their identity provider. Existing
deployments that registered `OpenCloudDesktop` should update the
client identifier on the IdP side to match, otherwise the audience
mismatch described above will continue to occur. Both spellings
can be registered in parallel during a transition window if desired.

## References

- RFC 6749, Section 2.2 (client identifier)
- OpenID Connect Core 1.0, Section 2 (`azp` claim) and
  Section 3.1.3.7 (ID Token validation)
- kanidm/kanidm#4473 — original report
- opencloud-eu/desktop#217 — possibly related, not resolved by
  this PR

### DIFF
--- a/src/libsync/theme.cpp
+++ b/src/libsync/theme.cpp
@@ -320,7 +320,7 @@ QIcon Theme::wizardHeaderLogo() const
 
 QString Theme::oauthClientId() const
 {
-    return QStringLiteral("OpenCloudDesktop");
+    return QStringLiteral("openclouddesktop");
 }
 
 QString Theme::oauthClientSecret() const
--- a/test/testoauth/testoauth.cpp
+++ b/test/testoauth/testoauth.cpp
@@ -274,11 +274,11 @@ public:
           ],
           "at_hash": "jEL4ptHeYx4eQa847tOVoQ",
           "aud": [
-            "OpenCloudDesktop"
+            "openclouddesktop"
           ],
           "auth_time": 1737560752,
-          "azp": "OpenCloudDesktop",
-          "client_id": "OpenCloudDesktop",
+          "azp": "openclouddesktop",
+          "client_id": "openclouddesktop",
           "email": "admin@admin.admin",
           "email_verified": true,
           "exp": 1739884152,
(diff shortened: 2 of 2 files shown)


yaleman:
Reminder that https://github.com/opencloud-eu/desktop/issues/246 would avoid this class of problem

TheOneRing:
I assume you have tested your change thoroughly with all existing idps and the supported deployments of OpenCloud....

I expect this change to be breaking for most existing setups.

roder:
No, I don't have the infrastructure or test harnesses to do that. I assumed
you did.  I also expect this to be a breaking change... but it also doesn't
look like a lot of users can use OIDC for auth either - so it's already
broke, right?

TheOneRing:
We have many working setups out there, just with other idps.

Anyhow, the way to go is to use the now merged web finger approach of https://github.com/opencloud-eu/desktop/pull/847

PR #985, opened 2026-07-21
test(gui): fix flaky spaces test suite

saw-jan:
- Catch the webdriver error and treat it as element not found. There can be an error even if the element is visible due to reload or refresh.
- split gui test suites

### DIFF
--- a/.woodpecker/ui-tests.yaml
+++ b/.woodpecker/ui-tests.yaml
@@ -36,9 +36,11 @@ workspace:
 matrix:
   include:
     - MATRIX_NAME: gui-tests-1
-      SUITES: 'features/activity features/add-account features/delete-files-folders features/edit-files features/login-logout features/move-files-folders'
+      SUITES: 'features/add-account features/delete-files-folders features/edit-files features/login-logout features/move-files-folders'
     - MATRIX_NAME: gui-tests-2
-      SUITES: 'features/remove-account-connection features/spaces features/sync-resources features/tabs-settings features/vfs'
+      SUITES: 'features/activity features/remove-account-connection features/spaces features/sync-edge-cases features/tabs-settings'
+    - MATRIX_NAME: gui-tests-3
+      SUITES: 'features/sync-resources features/vfs'
 
 steps:
   - name: restore-python-cache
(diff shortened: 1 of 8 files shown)


saw-jan:
^ @Asmitapaudel

.woodpecker/ui-tests.yaml
```diff
@@ -38,7 +38,9 @@ matrix:
     - MATRIX_NAME: gui-tests-1
       SUITES: 'features/activity features/add-account features/delete-files-folders features/edit-files features/login-logout features/move-files-folders'
     - MATRIX_NAME: gui-tests-2
-      SUITES: 'features/remove-account-connection features/spaces features/sync-resources features/tabs-settings features/vfs'
+      SUITES: 'features/remove-account-connection features/spaces features/sync-edge-cases features/tabs-settings'
+    - MATRIX_NAME: gui-tests-3
+      SUITES: 'features/sync-resources features/vfs'
```
prashant-gurung899, line 43: We could also add few more test suites here and distribute among pipelines equally.

test/gui/features/sync-edge-cases/syncResources.feature
prashant-gurung899, line 1: could we change the filename for this one?
Using same name will be confusing.

saw-jan: IMO, it should be fine. suite name will differentiate the type. The test scenarios are actually for syncing resources

.woodpecker/ui-tests.yaml
saw-jan: all runs for about ~10 minutes. looks pretty balanced for now. sync-resources suite has more scenarios

saw-jan: but may be we can add small suite form group 1

prashant-gurung899: Yeah, the run time for all groups seem to be around 10 minutes. So, we can keep it as it is for now.

prashant-gurung899:
LGTM 👍🏼

prashant-gurung899 (approved):

PR #986, opened 2026-07-22
chore(gui-test): additional ruff linter rules

Asmitapaudel:
* Configure Ruff with a `line-length` of **100**.
* Enable the following Ruff rule sets:
  * **`DTZ`** – Detect uses of naive `datetime` objects and encourage explicit timezone handling.
  * **`RET`** – Detect unnecessary or missing `return` statements and simplify return-related control flow.
* Re-enable **`E501`** by removing it from the global ignore list.
* Keep **`E501`** ignored only for `steps/*.py`, where long Behave step definitions are more readable and difficult to split.

### DIFF
--- a/test/gui/helpers/SpaceHelper.py
+++ b/test/gui/helpers/SpaceHelper.py
@@ -111,8 +111,9 @@ def create_space_file(space_name, file_name, content):
     response = request.put(url, content)
     if response.status_code not in (201, 204):
         raise AssertionError(
-            f"Creating file '{file_name}' in space '{space_name}' failed with {response.status_code}\n"
-            + response.text
+            f"Creating file '{file_name}' in space "
+            f"'{space_name}' failed with "
+            f"{response.status_code}\n"
         )
 
 
--- a/test/gui/helpers/StacktraceHelper.py
+++ b/test/gui/helpers/StacktraceHelper.py
@@ -41,7 +41,8 @@ def parse_stacktrace(coredump_file):
     message = []
     if coredump_file:
         coredump_filename = os.path.basename(coredump_file)
-        # example coredump file: core-1648445754-1001-11-!drone!src!build-GUI-tests!bin!opencloud
+        # example coredump file:
+        #       core-1648445754-1001-11-!drone!src!build-GUI-tests!bin!opencloud
         patterns = coredump_filename.split('-')
         app_binary = 'opencloud'
         if len(patterns) == 1:
@@ -52,7 +53,7 @@ def parse_stacktrace(coredump_file):
             app_binary = '-'.join(patterns[4:]).replace('!', '/')
 
         timestamp = datetime.fromtimestamp(
-            float(patterns[1] if patterns[1] != 'N/A' else time.time())
+            float(patterns[1] if patterns[1] != 'N/A' else time.time()), tz=datetime.UTC
         )
         message.append('-------------------------------------------')
         message.append(f'Executable: {app_binary}')
--- a/test/gui/helpers/SyncHelper.py
+++ b/test/gui/helpers/SyncHelper.py
@@ -29,7 +29,11 @@ else:
         content = content.replace('socketConnect = SocketConnect()', '')
         content = content.replace(
             'from gi.repository import GObject, Nautilus',
-            'import gi\n\ngi.require_version(\'Nautilus\', \'4.0\')\nfrom gi.repository import GObject, Nautilus',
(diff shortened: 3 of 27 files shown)


test/gui/helpers/SpaceHelper.py
```diff
@@ -111,8 +111,9 @@ def create_space_file(space_name, file_name, content):
     response = request.put(url, content)
     if response.status_code not in (201, 204):
         raise AssertionError(
-            f"Creating file '{file_name}' in space '{space_name}' failed with {response.status_code}\n"
-            + response.text
+            f"Creating file '{file_name}' in space "
+            f"'{space_name}' failed with "
+            f"{response.status_code}\n"
```
anon-pradip, line 116: This refactoring drops the server error details. Without it, test failures for file creation only show the HTTP status code with no debugging context (e.g., just 403 instead of 403 - "insufficient permissions").

test/gui/helpers/StacktraceHelper.py
```diff
@@ -52,7 +53,7 @@ def parse_stacktrace(coredump_file):
             app_binary = '-'.join(patterns[4:]).replace('!', '/')
 
         timestamp = datetime.fromtimestamp(
-            float(patterns[1] if patterns[1] != 'N/A' else time.time())
+            float(patterns[1] if patterns[1] != 'N/A' else time.time()), tz=datetime.UTC
```
anon-pradip, line 56: `datetime.UTC` was added in Python 3.11 ([PEP 615](https://docs.python.org/3/library/datetime.html#constants)), but project declares `requires-python = ">=3.10"`. So it won't break in practice, but the requires-python constraint needs updating.
Options:
- Update `requires-python = ">=3.11"` in `pyproject.toml` to match actual usage => This sounds better for me
- Or change to `datetime.timezone.utc` for widest compatibility

test/gui/helpers/SyncHelper.py
```diff
@@ -304,7 +303,7 @@ def wait_for_resource_to_sync(
                     + '\n'.join(sync_info)
                 )
         return
-    elif not force_sync:
+    if not force_sync:
```
anon-pradip, line 306: Can you please enlighten me about this change here?

test/gui/pageObjects/AccountSetting.py
```diff
@@ -86,9 +82,7 @@ def wait_until_account_is_connected(timeout=get_config('min_timeout')):
 
         if not result:
             raise TimeoutError(
-                "Timeout waiting for the account to be connected for "
-                + str(timeout)
-                + " seconds"
+                "Timeout waiting for the account to be connected for " + str(timeout) + " seconds"
```
anon-pradip, line 85: While we are still here, can we use f-string format?
`f"Timeout waiting for the account to be connected for {timeout} seconds"`

test/gui/pageObjects/AccountSetting.py
```diff
@@ -110,7 +104,5 @@ def account_removed():
 
         if not result:
             raise TimeoutError(
-                "Timeout waiting for account to be removed for "
-                + str(timeout)
-                + " seconds"
+                "Timeout waiting for account to be removed for " + str(timeout) + " seconds"
```
anon-pradip, line 107: same as above ^

saw-jan (approved):

test/gui/helpers/StacktraceHelper.py
saw-jan: this file is obsolete. we can remove this helper

test/gui/pageObjects/AccountSetting.py
saw-jan: CC @Asmitapaudel

test/gui/pageObjects/AccountSetting.py
saw-jan: CC @Asmitapaudel

test/gui/helpers/SyncHelper.py
saw-jan: CC @Asmitapaudel

test/gui/helpers/SpaceHelper.py
saw-jan: CC @Asmitapaudel

anon-pradip:
The above comments are addressed by https://github.com/opencloud-eu/desktop/pull/994. cc @saw-jan

Issue #809, opened 2026-02-15
Ability to use system icons


MorsMortium:
I saw https://github.com/opencloud-eu/desktop/issues/159 and the reasoning there for switching to FontAwesome, however on linux icons are part of what makes applications fit/blend into the desktop. For this reason I'd like to request an option to be able to use the system icons.

MorsMortium:
Really not productive

MorsMortium:
Yes, still an issue

MorsMortium:
Yes, still an issue

MorsMortium:
Yes, still an issue

TheOneRing:
Sorry, currently not planned.

PR #989, opened 2026-07-22
Implement optional icon coloring with text color

MorsMortium:
I have written the https://github.com/opencloud-eu/desktop/issues/809 issue originally.
This is a much simpler solution to my issue, which is the icon coloring.
It does not depend on system icons, so there is no possibility of one of them missing.
The restart mechanic is there as some parts of the GUI, mainly the top bar, do not refresh later. It was simpler this way, but if it should be avoided at all costs, I can look into adding a signal and refreshing when emitted.

### DIFF
--- a/src/gui/generalsettings.cpp
+++ b/src/gui/generalsettings.cpp
@@ -78,6 +78,11 @@ GeneralSettings::GeneralSettings(QWidget *parent)
     });
 
     connect(_ui->about_pushButton, &QPushButton::clicked, ocApp(), &Application::showAbout);
+
+    connect(_ui->textIconColorCheckBox, &QCheckBox::toggled, this, [this](bool checked) {
+        ConfigFile().setTextIconColor(checked);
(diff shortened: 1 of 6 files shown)


TheOneRing:
I'm sorry, but the colours are part of the branding, and thus we don't want then changed.

MorsMortium:
So for anyone else not wanting a broken theme, I have written a small LD_PRELOAD library: https://codeberg.org/MorsMortium/Setup/src/branch/main/OpenCloud.cpp
Here is how it is compiled: https://codeberg.org/MorsMortium/Setup/src/branch/main/UserSetup.sh#L109
And here is how it is applied: https://codeberg.org/MorsMortium/Setup/src/branch/main/UserSetup.sh#L159
This allows to unbreak the theme and strip the "branding", without ever needing to compile OpenCloud itself, resulting in a look like so (regarding icons):
image

TheOneRing:
You never stated that anything is broken, just that you didn't like it.
With a dark theme, the colour is supposed to be lilac, which should give you enough contrast to the dark colour of the theme.

MorsMortium:
It is not my theme color, so it is broken

PR #990, opened 2026-07-24
test(gui): retry getting clipboard data

saw-jan:
retry getting clipboard if it fails due to error and log the actual value on failure.

### DIFF
--- a/test/gui/pageObjects/AccountConnectionWizard.py
+++ b/test/gui/pageObjects/AccountConnectionWizard.py
@@ -1,6 +1,8 @@
 import os
+from time import time
 from types import SimpleNamespace
 from appium.webdriver.common.appiumby import AppiumBy as By
+from selenium.common.exceptions import WebDriverException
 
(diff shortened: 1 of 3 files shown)


saw-jan:
^ @Asmitapaudel

test/gui/pageObjects/AccountConnectionWizard.py
```diff
@@ -95,10 +97,25 @@ def copy_login_url():
             AccountConnectionWizard.COPY_URL_TO_CLIPBOARD_BUTTON.selector,
         ).click()
 
+    @staticmethod
+    def get_login_url():
+        try:
```
Asmitapaudel, line 103:
```suggestion
login_url = None
        try:
```
can we initialize  `login_url ` before the try? . If `copy_login_url()` fails before it's assigned, this will raise error and hide the original exception

Asmitapaudel:
LGTM

PR #993, opened 2026-07-27
refactor(gui-test): reduce Ruff ignores by fixing violations

Asmitapaudel:
### Summary

Reduced the number of ignored Ruff rules by fixing the corresponding lint violations in the codebase.

### Removed from ignore list

* `F403` – Wildcard import (`from x import *`)
* `E722` – Bare `except`
* `UP012` – Remove unnecessary UTF-8 encoding
* `SIM113` – Use `enumerate()` instead of a manual loop counter
* `N801` – Non-CapWords class names

### DIFF
--- a/test/gui/steps/sync_context.py
+++ b/test/gui/steps/sync_context.py
@@ -156,13 +156,11 @@ def step(context):
 @Then('the folders should be in the following order:')
 def step(context):
     row_index = 0
-    for row in context.table:
+    for row_index, row in enumerate(context.table):
         expected_folder = row[0]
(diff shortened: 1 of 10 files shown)


test/gui/steps/sync_context.py
```diff
@@ -156,14 +156,12 @@ def step(context):
 @Then('the folders should be in the following order:')
 def step(context):
     row_index = 0
```
saw-jan, line 158: this can be removed

saw-jan (approved):

PR #994, opened 2026-07-27
fix(gui-test): restore missing error details, remove dead code

anon-pradip:
Fix dropped server error details, remove dead `StacktraceHelper`, modernize `f-strings`. Leftover from https://github.com/opencloud-eu/desktop/pull/986


saw-jan (approved):

saw-jan (changes requested):
Could you address this as well?
https://github.com/opencloud-eu/desktop/pull/986/changes/BASE..8e4995ba6e326d88ce065a125591518a4aabe1f6#r3654575577

anon-pradip:
> Could you address this as well? https://github.com/opencloud-eu/desktop/pull/986/changes/BASE..8e4995ba6e326d88ce065a125591518a4aabe1f6#r3654575577

Used `if` instead of `elif` because Ruff's `RET505` explicitly flags `elif` after return as unnecessary: https://docs.astral.sh/ruff/rules/superfluous-else-return/
They say: _"The else statement is not needed as the return statement will always break out of the enclosing function. Removing the else will reduce nesting and make the code more readable."_

saw-jan (approved):

PR #995, opened 2026-07-27
chore(gui-test): add ARG ruff lint rule

Asmitapaudel:

### DIFF
--- a/test/gui/pyproject.toml
+++ b/test/gui/pyproject.toml
@@ -48,7 +48,11 @@ select = [
     "PLW2901",
     "DTZ",
     "RET",
+    "ARG",
 ]
+
+# Treat the following variable pattern as intentionally unused.
+dummy-variable-rgx = "^(_.*|context|feature|self)$"
 
 # See the available rules: https://docs.astral.sh/ruff/rules/
 ignore = [
--- a/test/gui/steps/account_context.py
+++ b/test/gui/steps/account_context.py
@@ -121,6 +121,7 @@ def step(context):
 
 @When('the user "{username}" logs out using the client-UI')
 def step(context, username):
+    Toolbar.open_account(username)
     AccountSetting.logout()
 
 
@@ -158,6 +159,7 @@ def step(context, _):
 
 @Then('user "{username}" should be connected to the server')
 def step(context, username):
+    Toolbar.open_account(username)
     AccountSetting.wait_until_account_is_connected()
 
 
--- a/test/gui/steps/file_context.py
+++ b/test/gui/steps/file_context.py
@@ -158,8 +158,8 @@ def step(context, username, foldername):
     create_folder(foldername, username)
 
 
-@When('user "{user}" creates a file "{filename}" with size "{filesize}" inside the sync folder')
-def step(context, user, filename, filesize):
+@When('user "{_user}" creates a file "{filename}" with size "{filesize}" inside the sync folder')
+def step(context, _user, filename, filesize):
     create_file_with_size(filename, filesize)
 
 
--- a/test/gui/Makefile
+++ b/test/gui/Makefile
@@ -47,7 +47,7 @@ run-atpsi-webdriver: check-uv
 .PHONY: python-lint
 python-lint: check-uv
 	uv run --no-sync ruff format . --diff
-	uv run --no-sync ruff check . --diff
+	uv run --no-sync ruff check .
 
 .PHONY: python-lint-fix
 python-lint-fix: check-uv
--- a/test/gui/pageObjects/AccountConnectionWizard.py
+++ b/test/gui/pageObjects/AccountConnectionWizard.py
@@ -4,7 +4,6 @@ from types import SimpleNamespace
 from appium.webdriver.common.appiumby import AppiumBy as By
 from selenium.common.exceptions import WebDriverException
 
-
 from helpers.WebUIHelper import authorize_via_webui
 from helpers.ConfigHelper import get_config
 from helpers.SetupClientHelper import (
@@ -69,7 +68,7 @@ class AccountConnectionWizard:
             AccountConnectionWizard.SERVER_ADDRESS_BOX.selector,
         )
         url_input.clear()
-        url_input.send_keys(get_config("localBackendUrl"))
+        url_input.send_keys(server_url)
 
         AccountConnectionWizard.next_step()
 
--- a/test/gui/pageObjects/Activity.py
+++ b/test/gui/pageObjects/Activity.py
@@ -88,7 +88,7 @@ class Activity:
                 return False
 
     @staticmethod
-    def select_synced_filter(sync_filter):
+    def select_synced_filter(_sync_filter):
         menu = app().find_element(
             Activity.LOCAL_ACTIVITY_FILTER_BUTTON.by,
             Activity.LOCAL_ACTIVITY_FILTER_BUTTON.selector,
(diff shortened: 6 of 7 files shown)


test/gui/steps/account_context.py
```diff
@@ -175,7 +175,7 @@ def step(context):
 @When('the user adds the server "{server_url}"')
 def step(context, server_url):
     server_url = substitute_inline_codes(server_url)
-    AccountConnectionWizard.add_server(server_url)
+    AccountConnectionWizard.add_server()
```
saw-jan: we are explicitly passing server url from here. This should not be changed. fix the `AccountConnectionWizard.add_server` instead.
```suggestion
    AccountConnectionWizard.add_server(server_url)
```

test/gui/steps/sync_context.py
```diff
@@ -273,7 +273,7 @@ def step(context):
 @When('the user checks the activities of account "{account}"')
 def step(context, account):
     account = substitute_inline_codes(account)
-    Activity.select_synced_filter(account)
+    Activity.select_synced_filter()
```
saw-jan: this should not be changed. We need this, the problem is that currently we cannot select the desired filter.

test/gui/steps/sync_context.py
```diff
@@ -297,8 +297,8 @@ def step(context):
 
 
 @When('the user unchecks the "{filter_option}" filter')
-def step(context, filter_option):
-    Activity.select_not_synced_filter(filter_option)
+def step(context, filter_option):  # noqa: ARG001
+    Activity.select_not_synced_filter()
```
saw-jan: should not be changed

test/gui/steps/file_context.py
```diff
@@ -159,7 +159,7 @@ def step(context, username, foldername):
 
 
 @When('user "{user}" creates a file "{filename}" with size "{filesize}" inside the sync folder')
-def step(context, user, filename, filesize):
+def step(context, user, filename, filesize):  # noqa: ARG001
```
saw-jan:
```suggestion
def step(context, _, filename, filesize):
```

test/gui/pyproject.toml
```diff
@@ -48,8 +48,12 @@ select = [
     "PLW2901",
     "DTZ",
     "RET",
+    "ARG",
 ]
 
+dummy-variable-rgx = "^(_.*|context|feature|self)$"  #Treat `context`, `feature`, `self`, and underscore-prefixed variables as intentionally unused.
+
+
```
saw-jan:
```suggestion
dummy-variable-rgx = "^(_.*|context|feature|self)$"  #Treat `context`, `feature`, `self`, and underscore-prefixed variables as intentionally unused.

```

test/gui/pyproject.toml
```diff
@@ -48,8 +48,12 @@ select = [
     "PLW2901",
     "DTZ",
     "RET",
+    "ARG",
 ]
 
+dummy-variable-rgx = "^(_.*|context|feature|self)$"  #Treat `context`, `feature`, `self`, and underscore-prefixed variables as intentionally unused.
```
saw-jan:
```suggestion
# Treat the following variable pattern as intentionally unused.
dummy-variable-rgx = "^(_.*|context|feature|self)$"
```

test/gui/steps/file_context.py
Asmitapaudel: I tried changing the parameter to _ as suggested, but Behave fails at runtime. https://ci.opencloud.rocks/repos/11/pipeline/516/59#L440

test/gui/helpers/AppHelper.py
```diff
@@ -64,7 +64,7 @@ def find_element(self, by, selector, timeout=None):
             set_implicit_wait(get_config('min_timeout'))
 
 
-def pause(self):
+def pause():
     threading.Event().wait()
```
anon-pradip: I see `pause()` is not called anywhere in the codebase - not as `app().pause()`, not as bare `pause()`, not anywhere. It is dead code. So we can remove this and its refs below too.

test/gui/steps/file_context.py
saw-jan: then we need to find other way. maybe try `_user`

test/gui/helpers/AppHelper.py
saw-jan: hmm, true but this method is an extension for webdriver for debugging purpose. this should not be removed.

test/gui/steps/file_context.py
Asmitapaudel: It also shows the same error. Can we change the step parameter from `user`  to ` _user`?

test/gui/helpers/AppHelper.py
```diff
@@ -64,13 +64,8 @@ def find_element(self, by, selector, timeout=None):
             set_implicit_wait(get_config('min_timeout'))
 
 
-def pause(self):
-    threading.Event().wait()
-
-
 # bind custom element methods
 Remote.find_element = find_element
-Remote.pause = pause
```
saw-jan, line 73: we need this for debugging

test/gui/steps/file_context.py
saw-jan: will changing to `_user` work?

test/gui/helpers/AppHelper.py
```diff
@@ -64,7 +64,7 @@ def find_element(self, by, selector, timeout=None):
             set_implicit_wait(get_config('min_timeout'))
 
 
-def pause(self):
+def pause():
```
saw-jan: will this work without self?

Asmitapaudel: It didn't throw any errors in CI. we're already ignoring `self`, can we restore it?

test/gui/steps/file_context.py
Asmitapaudel: yes,  if we change in a step parameter also then it works.

test/gui/helpers/AppHelper.py
saw-jan: what happens when you use it while running tests?

Asmitapaudel: for running the test it doesn't effect. it is not used in normal test . If driver.pause() is called for debugging, removing self could cause an issue.

test/gui/steps/account_context.py
```diff
@@ -120,7 +120,7 @@ def step(context):
 
 
 @When('the user "{username}" logs out using the client-UI')
-def step(context, username):
+def step(context, username):  # noqa: ARG001
     AccountSetting.logout()
```
saw-jan: we can use the username like this
```suggestion
def step(context, username):
    Toolbar.open_account(username)
    AccountSetting.logout()
```

test/gui/steps/account_context.py
```diff
@@ -157,7 +157,7 @@ def step(context, _):
 
 
 @Then('user "{username}" should be connected to the server')
-def step(context, username):
+def step(context, username):  # noqa: ARG001
     AccountSetting.wait_until_account_is_connected()
```
saw-jan:
```suggestion
def step(context, username):
    Toolbar.open_account(username)
    AccountSetting.wait_until_account_is_connected()
```

test/gui/behave.ini
```diff
@@ -1,6 +1,6 @@
 [behave]
 paths=features
-default_format = pretty
+default_format = plain
```
saw-jan: ⚠️

saw-jan (approved):

PR #1010, opened 2026-08-05
Disable HTTP/2 support again

### LINKED ISSUE #734: 3.0 - Server dislikes our behavior, excessive load detected.
Setup: 
- Server: 4.0-rc1 & 4.0-rc2 
- Client 1: ArchLinux/3.0.1 - Build from Source (AUR)
- Client 2: macOS/3.0.2
- Client 3: Windows/3.0.2 (Unaffected)
- Proxy: nginx

With the new client, I encounter the problem that the synchronization stops and I receive the error “Server dislikes our behavior, excessive load detected.” Which seems to be related to QT (https://codebrowser.dev/qt5/qtbase/src/network/access/http2/http2protocol.cpp.html). 

I’m using the recommend settings from https://docs.opencloud.eu/docs/admin/getting-started/container/docker-compose/external-proxy. 

The issue doesn’t occur with the 2.0 client (Linux/AppImage) 

Within the nginx and container logs, I was unable to find any specific error message related to the issue (default log level).

**Windows**: I'm unable to reproduce the issue with the Windows Client. I've made the folder (6.500 Files / 65 GB), that I've used for testing, offline available and everything went fine. Under Linux and macOS the process has to be restarted every few minutes.

TheOneRing:
Release 3.0 showed that Qt is still not ready

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -102,6 +102,15 @@ option(BETA_CHANNEL_BUILD "Use a standalone profile for a beta release" OFF)
 ## -DFORCE_ASSERTS=ON
 option(FORCE_ASSERTS "FORCE_ASSERTS" OFF)
 
+## -DWITH_HTTP2=ON
+## For testing purposes the client behavior can also be modified with an environment variable OPENCLOUD_USE_HTTP2=[0|1]
+## Relevant bugs:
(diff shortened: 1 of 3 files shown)


guruz (approved):
Looks good..

FYI there was a comment on the qt upstream issue yesterday, so maybe in a few weeks we can re-enable again.

PR #1024, opened 2026-08-13
Mac signing workflow

flimmy:
add mac signing and the respective secret handling to the workflow.
The secret handling is supposed to be reusable.


.github/workflows/craft_override.ini
```diff
@@ -14,6 +14,17 @@ CodeSigning/Enabled = True
 CodeSigning/CommonName = Test certificate for 'OpenCloud Desktop Client [OSS]'
 CodeSigning/WindowsCustomSignCommand = pwsh -File "${Env:GITHUB_WORKSPACE}/.github/workflows/.signpath.ps1" -F "%F"
 
+[macos-clang-x86_64]
+CodeSigning/Enabled = True
+CodeSigning/MacKeychainPath = ${Env:SIGNING_KEY_FILE_PATH}
+Packager/PackageType = MacPkgPackager
+
+[macos-clang-arm64]
+CodeSigning/Enabled = True
```
TheOneRing, line 24: I guess we could also the keychain path for other platforms as the value would not be used.
So it would reduce the duplication.

You can also set default values in https://github.com/opencloud-eu/desktop/blob/main/.craft.ini#L103 which kick in if there is no env var.
Which could reduce the complexity of the env handling in the main.yml

.github/workflows/main.yml
```diff
@@ -202,13 +270,18 @@ jobs:
...
+        Remove-Item -force "${{ matrix.signing.keyfile_file_path }}\${{ matrix.signing.keyfile_file_name }}"
+
     - name: Package msix
       if: ${{ startsWith(matrix.os, 'windows-') && github.event_name != 'pull_request' }}
       run: |
         & "${env:GITHUB_WORKSPACE}/.github/workflows/.craft.ps1" -c --no-cache --package --options "[Packager]PackageType=AppxPackager" --options "[Packager]Destination=${{ github.workspace }}/msix/" opencloud/opencloud-desktop
         Move-Item -Path "${{ github.workspace }}/msix/*.msix" -Destination "${{ github.workspace }}/msix/OpenCloud_Desktop-${{ github.run_number }}.msix" -Force
 
-    - name: Upload binary artifacts
+    - name: Upload binary binary artifacts
```
TheOneRing: binary binary.
Pika pika

TheOneRing (approved):
looks good.
and yay many thanks, finally usable artifacts

.github/workflows/main.yml
```diff
@@ -129,6 +154,8 @@ jobs:
             & "${env:GITHUB_WORKSPACE}/.github/workflows/.craft.ps1" -c dev-utils/linuxdeploy
             # ensure we have fuse3-devel installed
             dnf install -y fuse3-devel file
+        } elseif($IsMacOS) {
```
TheOneRing: can be skipped if we don't package

flimmy: we can package even if we don't sign if https://github.com/opencloud-eu/desktop/pull/1025 is implemented

flimmy: skipped for mac if no signing

.github/workflows/craft_override.ini
```diff
@@ -14,6 +16,14 @@ CodeSigning/Enabled = True
 CodeSigning/CommonName = Test certificate for 'OpenCloud Desktop Client [OSS]'
 CodeSigning/WindowsCustomSignCommand = pwsh -File "${Env:GITHUB_WORKSPACE}/.github/workflows/.signpath.ps1" -F "%F"
 
+[macos-clang-x86_64]
+CodeSigning/Enabled = True
+Packager/PackageType = MacPkgPackager
+
+[macos-clang-arm64]
+CodeSigning/Enabled = True
```
TheOneRing, line 24: In a later step `$env:SIGN_PACKAGE` is used. Which would already enable code signing (`CodeSigning/Enabled = ${Env:SIGN_PACKAGE}` in the main config.
However, I can't see that `${Env:SIGN_PACKAGE}` is set anywhere?

flimmy: i used the same override the windows step uses.
The SIGN_PACKAGE env is set to falls in the .craft.ini not the override.

But nevertheless the next pr https://github.com/opencloud-eu/desktop/pull/1025/changes will remove the explicit CodeSigning/Enabled override and move it into the environment using the SIGN_PACKAGE env-var

TheOneRing (approved):

PR #1027, opened 2026-08-17
Fix issue with unicode paths in qml

### LINKED ISSUE #938: [Bug] Error explorer not found
## Steps to reproduce
1. latest desktop on win 11
2. click on show in explorer
3. see error

## Expected behavior
- should open explorer

Image
Image

TheOneRing:
Instead of constructing a file path in qml we now expose and use showInFileManager in qml

Fixes: #938


guruz (commented):
👍

PR #1028, opened 2026-08-17
Retry PROPFIND/GET/HEAD on a status-0 transport failure instead of aborting sync

### LINKED ISSUE #1026: Sync aborts the whole run when a PROPFIND returns HTTP status 0 with an empty error (client-side HTTP/2 GOAWAY)
## Summary

During discovery of a large tree, single PROPFIND requests come back to the client as transport-level failures: HTTP status `0`, empty header object, zero-length body, and no `error` key at all. The desktop client treats such a reply as a fatal server error: in this run, it aborted the entire sync in the same millisecond rather than retrying that single request.

What I consistently observed immediately before each abort is an HTTP/2 `GOAWAY` on the client-to-proxy hop. Qt's own `qt.network.http2` log category recorded such a line immediately before the aborted run in each of the three runs where I had the `qt.network.http2` category enabled. Every server-side log line for the affected runs is `"proto":"HTTP/1.1"`, because the reverse proxy speaks HTTP/1.1 upstream, so the HTTP/2 hop where the GOAWAY occurs is client to proxy, and the handling of the resulting status-0 reply is client side.

The request I am making here is narrow and is about the client's reaction, not about Qt's HTTP/2 stack: a reply with HTTP status `0` and an empty error string is not a server error and should not abort a whole sync run.

## Environment

- Client: OpenCloud Desktop, Windows, MSIX package version `3.3.2073.0`. The client's own About banner reports a different version string: `OpenCloud 3.0.3.2073 72fe8f Qt Bibliotheken 6.9.2, OpenSSL 3.5.4 30 Sep 2025 ... cfapi OS: windows-10.0.26200`. User-Agent sent is `mirall/3.0.3.2073`. I am quoting both rather than picking one, since the package version and the internal version do not match.
- Server: OpenCloud 7.2.3.
- Reverse proxy: Nginx Proxy Manager. HTTP/2 exists only on the client-to-proxy hop; the proxy talks `proxy_http_version 1.1` to the upstream, confirmed by `"proto":"HTTP/1.1"` on every server-side log line for the affected runs.
- Sync target: a space with a large tree (over 4000 paths).

## Steps to reproduce (as observed)

1. Connect the Windows desktop client to an OpenCloud server that sits behind an HTTP/2-terminating reverse proxy.
2. Sync a space large enough that a full discovery issues several thousand PROPFINDs.
3. Let a full discovery run.

Observed: the run was aborted and discovery did not complete. On my setup this happened on every full-discovery burst until I changed a proxy-side connection setting (see Evidence, control run). I am deliberately not asserting an exact numeric trigger or where in the tree any given attempt stopped, because I could not measure per-connection request counts and have no data on abort position (see "What I could not verify").

## Expected behavior

A reply that carries HTTP status `0` and no error string is a transport-level failure of a single request, not a server-reported error condition. It should not be classified as a fatal server error for the directory, and it should not abort the whole sync run. At minimum, such a reply should be retried at the request level.

## Actual behavior

The status-0 reply is passed into discovery as a server error carrying HTTP status `0`, and the engine aborts the entire run immediately. The sync failed and did not recover on its own during the observed window.

### 1. The replies themselves

Recovered `sync.httplogger` RESPONSE payloads for five affected replies (request IDs starting `6a982df9`, `6c597349`, `2f52d868`, `868affb2`, `35499dbc`) all show status `0`, an empty `header` object, a zero-length body, and no `error` key. Two full, verbatim payloads (space UUID redacted as `<space-id>`):

```json
{"response":{"body":{"length":0},"header":{},"info":{"id":"6a982df9-3290-4f40-b5ba-fef2751c8f81","method":"PROPFIND","reply":{"cached":false,"duration":1,"durationString":"duration(0h, 0min, 0s, 1ms)","status":0,"version":"HTTP 2"},"url":"https://opencloud.bottich.lol/dav/spaces/<space-id>/designia/ressourcen/document_templates/powerpoint/Kebo PPT"}}}
```

```json
{"response":{"body":{"length":0},"header":{},"info":{"id":"35499dbc-dbe7-4d05-aab1-d9f0eca5b7ea","method":"PROPFIND","reply":{"cached":false,"duration":4,"durationString":"duration(0h, 0min, 0s, 4ms)","status":0,"version":"HTTP 2"},"url":"https://opencloud.bottich.lol/dav/spaces/<space-id>/designia/ressourcen/icons/icons-drweb"}}}
```

### 2. The abort is immediate and whole-run, not per-request

```
26-08-13 15:58:47:961 [ warning qt.network.http2 ]: stream 2001 error: "Received GOAWAY"
26-08-13 15:58:47:961 [ warning qt.network.http2 ]: stream 2003 error: "Received GOAWAY"
26-08-13 15:58:47:961 [ warning qt.network.http2 ]: stream 2005 error: "Received GOAWAY"
26-08-13 15:58:47:963 [ warning sync.discovery ]: Server error in directory "designia/ressourcen/document_templates/cv/the_green_cv" 0
26-08-13 15:58:47:963 [ info sync.engine ]: Aborting sync
26-08-13 15:58:47:963 [ warning gui.folder ]: SyncEngine finished with ERROR
```

The trailing ` 0` matches the status `0` seen in the recovered payloads; I read it as the HTTP status, though I did not trace this through source (see "What I could not verify"). Two milliseconds after the first GOAWAY line, the whole run is gone.

I had the `qt.network.http2` category enabled for three consecutive runs, and each one shows at least one such GOAWAY line before its abort. The specific streams affected varied: 2001/2003/2005 in the run shown above, 2001 only in the second run, and 2001 through 2013 (2001/2003/2005/2007/2009/2011/2013) in the third. Stream 2001 appears in all three runs; the three lines in the run shown above share the same millisecond timestamp, so I cannot say from the log which one the client actually processed first.

### 3. Before/after control

The only variable I changed was the proxy's `keepalive_requests`.

Before: every full-discovery burst from this client ended at one of these totals: 1001, 1001, 1001, 1003, 2008, 4007, 4011, 4021, 5005, 5006, 6004, 8005. No burst completed cleanly during the observed window.

After raising nginx's `keepalive_requests` from the default 1000 to 100000, one full discovery completed cleanly: 4068 PROPFINDs over 4068 distinct paths (ratio exactly 1.00), zero abandoned in-flight requests, space root requested exactly once.

I am reporting this as a correlation from a controlled single-variable change, not as proof of the exact mechanism.

Whether this relates to opencloud-eu/reva#767 is for a maintainer to judge.

## Impact

I would expect any mid-flight close of the client-to-proxy connection to produce the same status-0 reply, though `keepalive_requests` is the only such cause I tested. On my setup, no full-discovery burst completed cleanly during the observed window until that setting was changed. The run ends with `gui.folder: SyncEngine finished with ERROR`, and the reply itself carries no status text and no error string.

## Suggested fix

In the client, do not treat a reply with HTTP status `0` and an empty error string as a fatal, whole-run-aborting server error. Such a reply is a transport-level failure of one request. At minimum, retry it at the request level rather than aborting the whole sync.

I am deliberately not suggesting a change inside Qt's HTTP/2 handling, since that is Qt's component and not something this repository can implement.

## What I could not verify

- **GOAWAY frame details.** Neither the error code nor the Last-Stream-ID of the GOAWAY frame was captured. I therefore cannot say whether the affected streams were processed by the peer or not. Server-side evidence actually shows that some requests at the same boundary were processed (499/502 after up to 6.6s in flight), which argues against a simple "never processed" reading.
- **Per-connection request count.** No per-connection request count was ever measured; the nginx `log_format` in use has no `$connection` or `$connection_requests`. The stream numbers in the Qt log lines are reported as-is; I am not converting them into a request ordinal.
- **Determinism.** One separate run crossed a similar boundary without hitting this failure mode. I cannot say whether the failure is fully deterministic or timing dependent, and I am not resolving that either way.
- **Source-level confirmation.** I did not trace this through libsync. Everything above is from client logs, server logs, and one controlled proxy-side change. Whether a request-level retry or connection recycling path already exists and simply was not taken here, I cannot tell from logs alone.

Happy to provide fuller logs (with `qt.network.http2` and `sync.httplogger` enabled) if that helps.

junkerderprovinz:
## Summary

Fixes #1026. A PROPFIND reply with HTTP status `0` and no error text never reached the application layer - the connection was torn down before any response line was parsed (e.g. a client-side HTTP/2 GOAWAY racing an in-flight stream, as documented in #1026 with full client/server log evidence).

`discovery.cpp`'s `startAsyncServerQuery` handler currently classifies **any** such reply as a fatal, whole-run-aborting server error: it checks `code == 404` for the root path and `code >= 403` for a subdirectory (to gracefully ignore just that directory), but a status of `0` matches neither, so it always falls through to the final `fatalError` branch meant for "the root job... or network errors."

## Fix

`AbstractNetworkJob` already has a retry mechanism (`JobQueue`/`needsRetry()`) that recovers from specific transient conditions - unsupported redirects, expired auth, and (for HTTP/2 specifically) `ContentReSendError`. This PR extends `needsRetry()` with one more case: when `httpStatusCode() == 0` (no HTTP status was ever received) **and** the request verb is read-only/idempotent (`GET`, `HEAD`, `PROPFIND`). Since the server never confirmed it saw the request, retrying is safe for these verbs without the double-processing risk a write (`PUT`/`DELETE`/`MKCOL`/`MOVE`) would carry, so the verb check is intentionally conservative.

This means a single dropped PROPFIND during discovery gets retried (up to the existing `MaxRetryCount = 5`) exactly like the other already-handled recoverable cases, instead of aborting the entire sync run.

## Testing

Added `testRemoteDiscoveryRetriesTransportFailure` in `test/testremotediscovery.cpp`, modeled on the existing `testRemoteDiscoveryError` fixture: the fake server returns a status-0 `FakeErrorReply` for the first PROPFIND on a subdirectory, then a normal reply on retry, and asserts the sync completes with no `syncError` and the directory fully matches between local and remote.

I wasn't able to build/run this locally (no Qt/CMake toolchain available in my environment) - relying on CI to validate compilation and the new test. Happy to iterate on feedback.

## Note on PR #1010

#1026 was closed as a duplicate referencing #1010 ("Disable HTTP/2 support again"), which works around the whole class of GOAWAY-related failures by disabling HTTP/2 client-side. This PR is a narrower, independent fix for the underlying error-handling gap (any transport failure with no HTTP status aborts the whole sync, not just HTTP/2-specific ones) and should be safe to land alongside or independently of that workaround.

### DIFF
--- a/src/libsync/abstractnetworkjob.cpp
+++ b/src/libsync/abstractnetworkjob.cpp
@@ -129,6 +129,16 @@ bool AbstractNetworkJob::needsRetry() const
             }
             [[fallthrough]];
         default:
+            // A reply with no HTTP status at all (HttpStatusCodeAttribute unset, so
+            // httpStatusCode() reads back as 0) never reached the application layer - the
+            // connection was torn down before any response line was parsed, e.g. a
+            // client-side HTTP/2 GOAWAY racing an in-flight stream (see #1026). The server
+            // never confirmed it saw the request, so blindly retrying is only safe for
+            // read-only, idempotent verbs.
+            if (httpStatusCode() == 0
+                && (_verb == QByteArrayLiteral("GET") || _verb == QByteArrayLiteral("HEAD") || _verb == QByteArrayLiteral("PROPFIND"))) {
+                return true;
+            }
             break;
         }
     }
--- a/test/testremotediscovery.cpp
+++ b/test/testremotediscovery.cpp
@@ -178,6 +178,38 @@ private Q_SLOTS:
         QCOMPARE(errorSpy[0][0].toString(), formatAbortError(syncRootError.isEmpty() ? QString(fatalErrorPrefix + expectedErrorString) : syncRootError));
     }
 
+    // Regression test for #1026: a PROPFIND reply with HTTP status 0 and no error text
+    // (e.g. a client-side HTTP/2 GOAWAY racing an in-flight stream) is a transport
+    // failure, not a server error, and must be retried at the request level instead of
+    // aborting the whole sync run.
+    void testRemoteDiscoveryRetriesTransportFailure()
+    {
+        QFETCH_GLOBAL(Vfs::Mode, vfsMode);
+        QFETCH_GLOBAL(bool, filesAreDehydrated);
+
+        FakeFolder fakeFolder(FileInfo::A12_B12_C12_S12(), vfsMode, filesAreDehydrated);
+        fakeFolder.remoteModifier().insert(QStringLiteral("B/z2"));
+
+        QString errorFolder = Utility::concatUrlPath(OCC::TestUtils::dummyDavUrl(), QStringLiteral("B")).path();
+        bool firstAttempt = true;
+        fakeFolder.setServerOverride([&](QNetworkAccessManager::Operation op, const QNetworkRequest &req, QIODevice *) -> QNetworkReply * {
+            if (req.attribute(QNetworkRequest::CustomVerbAttribute).toByteArray() == "PROPFIND" && req.url().path().endsWith(errorFolder)) {
+                if (firstAttempt) {
+                    firstAttempt = false;
+                    // status 0, no error text: exactly what a client-side HTTP/2 GOAWAY produces
+                    return new FakeErrorReply(op, req, this, 0);
+                }
+            }
+            return nullptr;
+        });
+
+        QSignalSpy errorSpy(&fakeFolder.syncEngine(), &SyncEngine::syncError);
+        QVERIFY(fakeFolder.applyLocalModificationsAndSync());
+        QCOMPARE(errorSpy.size(), 0);
+        QVERIFY(!firstAttempt); // the transient reply was actually hit, not skipped
+        QCOMPARE(fakeFolder.currentRemoteState().children[QStringLiteral("B")], fakeFolder.currentLocalState().children[QStringLiteral("B")]);
+    }
+
     void testMissingData()
     {
         QFETCH_GLOBAL(Vfs::Mode, vfsMode);


TheOneRing:
We disable http2 all together in the upcoming release as there is a known Qt bug.
Therefore I don't think we need this change.

PR #1032, opened 2026-08-18
fix and streamline signing toggle variables

flimmy:
move the big switch-case from the environment selection into the first matrix variables (matrix.signing.enabled) as a yaml-anchor to make it reusable.
Decide everything else (signing related) only based on this (matrix.signing.enabled) boolean.
This also makes the workflow more transparent, as now the github deployment environment only contains the 3 bao_secrets and no other env-vars.


TheOneRing:
That syntax is way beyond my understanding of github actions...

TheOneRing (approved):

PR #1034, opened 2026-08-19
Properly shut down linux folder watcher

TheOneRing:


guruz (approved):
Looks good I guess, I am not an expert on Linux (or anything)...

Was there an issue that this fixes? Add to commit?

The IN_CLOEXEC is not necessary in our app, is it? But probably fine to have it.

TheOneRing:
Not an issue but in a recent review it was remarked that we don't shut them down.
As it is entirely possible that a sync is removed we should really remove the watcher in that case.

guruz:
Could be, no idea.

I would have thought the QScopedPointer does this implicitly on destruction.

TheOneRing:
Yes, the notifier was closed. But the file descriptor wasn't.
At least there is nothing the like in https://doc.qt.io/qt-6/qsocketnotifier.html#dtor.QSocketNotifier

Issue #843, opened 2026-03-11
VFS: Hydrated files marked as excluded


TheOneRing:
Hidden files and files that cannot be synced are marked as excluded.
When they are later renamed, no longer hidden, they keep the state excluded.

OpenVFS is affected, cfapi needs testing

TheOneRing:
Fixed in https://github.com/opencloud-eu/desktop/pull/1036

PR #1036, opened 2026-08-19
Addresses #843 - clean up the Excluded tag on VFS status change

NeoTheFox:
Add a SyncFileStatus condition to clean up the Excluded tag + a simple test
Tested on my local Arch machine ( I'm unable to test this on other operating systems)


TheOneRing:
Thx!

PR #1039, opened 2026-08-20
Cleanup exclude handling

TheOneRing:


guruz (approved):
It's ok I guess, but I would not change the test cases, you never know.

In general @dragotin is a better person for this one to review...

PR #1043, opened 2026-08-24
Propagate folder pin state to child items

NeoTheFox:
Add a poropogation test for OpenVFS

This addresses #844

Tested on my Arch system, I can't test it on OSX or Windows sadly.

Hopefully this plus my previous work could help the 4.0.0 milestone along.

### DIFF
--- a/src/plugins/vfs/openvfs/vfs_openvfs.cpp
+++ b/src/plugins/vfs/openvfs/vfs_openvfs.cpp
@@ -62,18 +62,18 @@ OCC::FileSystem::Path openVFSConfigFilePath()
     return {};
 }
 
-OpenVFS::PlaceHolderAttributes placeHolderAttributes(const std::filesystem::path &path)
+OpenVFS::PlaceHolderAttributes placeHolderAttributes(const std::filesystem::path &path, bool warnOnMissing = true)
 {
     const auto data = OCC::FileSystem::Xattr::getxattr(path, QString::fromUtf8(OpenVFS::Constants::XAttributeNames::Data));
-    if (!data) {
+    if (!data && warnOnMissing) {
         qCWarning(lcOpenVFS) << u"No OpenVFS xattr found for" << path.native();
     }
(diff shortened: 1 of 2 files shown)


TheOneRing:
Hmm yes, I created the issue, more like a note to myself. Sadly, I did not provide much context about my intensions back then...

But is that really needed, shouldn't we rather use "inherited" by default and look up a parent value if needed?
Setting it recursively on all children will become a very expansive operation....

NeoTheFox:
I thought that was a necessary addition before 4.0.0 is greenlit!
Well, all I did it match cfapi behavior here. 
Correct me if I'm wrong, that's just my second commit, but Inherited never gets checked (https://github.com/opencloud-eu/desktop/blob/b4c9826d26fe904e0fc1a0007d2693198f1a80bd/src/plugins/vfs/openvfs/vfs_openvfs.cpp#L498) 
So if this would be too much of a performance hit here I'm not sure how major of a rewrite a proper implemention would be.

TheOneRing:
> I thought that was a necessary addition before 4.0.0 is greenlit! Well, all I did it match cfapi behavior here. Correct me if I'm wrong, that's just my second commit, but Inherited never gets checked (
> 
> https://github.com/opencloud-eu/desktop/blob/b4c9826d26fe904e0fc1a0007d2693198f1a80bd/src/plugins/vfs/openvfs/vfs_openvfs.cpp#L498
> )
> So if this would be too much of a performance hit here I'm not sure how major of a rewrite a proper implemention would be.

Hmm your right the imported cfapi code does the same, not ideal either....
Yes it was previously a blocker for the 4.0 release, however due to a rather big blocker in openvfs 4.0 will be shipped without openvfs.
The milestone was just not... well maintained...

Current blockers for 4.0 are currently internal stuff like qa and code signing....

NeoTheFox:
Ah, ok. If I'll have some more time to look this over I'll try to come up with a better idea, but I was kinda relying completely on parroting cfapi. I'm not in a rush for this to be merged so we can just keep this as a WIP.

TheOneRing:
Well your pr made it clear to me that I have to think about how we handle this in general.....

PR #1051, opened 2026-08-26
CI: Build nightlies in beta configuration

TheOneRing:


flimmy (commented):
Ref type on nightlies is always branch. So the first part of the if should already trigger, making this change unnecessary. If it doesn't work already something is funky with the if

TheOneRing:
Your right, I didn't check the content of the artifacts. 🙈

PR #1059, opened 2026-08-27
test(gui): fix teardown and improve logs

saw-jan:
- Fix test teardown
- Abort client config creation upon failure
- Improve request assertion logs
- use `OC_LDAP_LOOKUP_CACHE_TTL=0` to fix race issue in the tests

### DIFF
--- a/test/gui/behave.ini
+++ b/test/gui/behave.ini
@@ -12,7 +12,7 @@ logging_level = WARNING
 capture = false
 capture_hooks = false
 show_skipped = false
-; Early fail. Stop the test run on first failure.
+; Config to fail early. Stop the test run on first failure.
 stop = false
 
 [behave.formatters]
--- a/test/gui/behave.ini
+++ b/test/gui/behave.ini
@@ -12,6 +12,8 @@ logging_level = WARNING
 capture = false
 capture_hooks = false
 show_skipped = false
+; Early fail. Stop the test run on first failure.
+stop = false
 
 [behave.formatters]
 html-pretty = behave_html_pretty_formatter:PrettyHTMLFormatter
--- a/test/gui/environment.py
+++ b/test/gui/environment.py
@@ -7,9 +7,9 @@ from helpers.ConfigHelper import init_config, reset_sync_connection_name
 from helpers.api.provisioning import delete_created_users
 from helpers.SpaceHelper import delete_project_spaces
 from helpers.ConfigHelper import get_config
-from helpers.FilesHelper import prefix_path_namespace, cleanup_created_paths
+from helpers.FilesHelper import cleanup_created_paths
 from helpers.AppHelper import close_and_kill_app
-from helpers.SyncHelper import clear_socket_messages
+from helpers.SyncHelper import clear_socket_messages, close_socket_connection
 from helpers.ReportHelper import (
     normalize_scenario_title,
     hit_screenrecord_limit,
@@ -44,6 +44,10 @@ def after_scenario(context, scenario):
 
     # quit the application
     close_and_kill_app()
+    clear_socket_messages()
+    close_socket_connection()
+    cleanup_current_app_log()
+    reset_sync_connection_name()
 
     # store app log on scenario failure
     if scenario.status in [Status.failed, Status.error] and os.path.exists(
@@ -56,22 +60,9 @@ def after_scenario(context, scenario):
 
     # clean up sync dir
     if os.path.exists(get_config("clientRootSyncPath")):
-        for entry in os.scandir(get_config("clientRootSyncPath")):
-            try:
-                if entry.is_file() or entry.is_symlink():
-                    print("Deleting file: " + entry.name)
-                    os.unlink(prefix_path_namespace(entry.path))
-                elif entry.is_dir():
-                    print("Deleting folder: " + entry.name)
-                    shutil.rmtree(prefix_path_namespace(entry.path))
-            except OSError as e:
-                print(f"Failed to delete '{entry.name}'.\nReason: {e}.")
-    # cleanup paths created outside of the temporary directory during the test
+        shutil.rmtree(get_config("clientRootSyncPath"))
+    # cleanup files and folders created during the test
     cleanup_created_paths()
+
     delete_project_spaces()
     delete_created_users()
-
-    cleanup_current_app_log()
-    clear_socket_messages()
-
-    reset_sync_connection_name()
--- a/test/gui/environment.py
+++ b/test/gui/environment.py
@@ -46,8 +46,6 @@ def after_scenario(context, scenario):
     close_and_kill_app()
     clear_socket_messages()
     close_socket_connection()
-    cleanup_current_app_log()
-    reset_sync_connection_name()
 
     # store app log on scenario failure
     if scenario.status in [Status.failed, Status.error] and os.path.exists(
@@ -61,8 +59,10 @@ def after_scenario(context, scenario):
     # clean up sync dir
     if os.path.exists(get_config("clientRootSyncPath")):
         shutil.rmtree(get_config("clientRootSyncPath"))
-    # cleanup files and folders created during the test
+
     cleanup_created_paths()
+    cleanup_current_app_log()
+    reset_sync_connection_name()
 
     delete_project_spaces()
     delete_created_users()
--- a/test/gui/helpers/SetupClientHelper.py
+++ b/test/gui/helpers/SetupClientHelper.py
@@ -91,59 +91,67 @@ def start_client():
 
 def generate_account_config(users, space='Personal'):
     sync_paths = {}
-    settings = QSettings(get_config('clientConfigFile'), QSettings.Format.IniFormat)
     users_uuids = {}
     server_url = get_config('localBackendUrl')
     capabilities = provisioning.get_capabilities()
-    capabilities_variant = QJsonValue(capabilities).toVariant()
 
-    for idx, username in enumerate(users):
-        users_uuids[username] = QUuid.createUuid()
-        settings.beginGroup("Accounts")
-        settings.beginWriteArray(str(idx + 1), len(users))
+    try:
+        settings = QSettings(get_config('clientConfigFile'), QSettings.Format.IniFormat)
+        capabilities_variant = QJsonValue(capabilities).toVariant()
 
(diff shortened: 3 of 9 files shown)


saw-jan:
^@Asmitapaudel

test/gui/environment.py
```diff
@@ -44,6 +44,10 @@ def after_scenario(context, scenario):
 
     # quit the application
     close_and_kill_app()
+    clear_socket_messages()
+    close_socket_connection()
+    cleanup_current_app_log()
```
Asmitapaudel: `cleanup_current_app_log()` removes `currentAppLogFile` with `os.remove()`. Since `save_app_log(scenario)` is called afterwards and checks whether the file exists, the log will already have been deleted when the scenario fails.

Could we move `save_app_log(scenario)` before `cleanup_current_app_log()` so the failure log is saved before the temporary log is removed?

test/gui/helpers/SyncHelper.py
```diff
@@ -164,6 +164,16 @@ def clear_socket_messages(resource=''):
...
 
+def close_socket_connection():
+    socket_messages.clear()
+    if SOCKET_CONNECT:
+        SOCKET_CONNECT.connected = False
+        if is_windows():
+            SOCKET_CONNECT.close_conn()
+        elif is_linux():
+            SOCKET_CONNECT._sock.close()
+
```
Asmitapaudel, line 182:
```suggestion
def close_socket_connection():
   global SOCKET_CONNECT
    socket_messages.clear()
    if SOCKET_CONNECT:
        SOCKET_CONNECT.connected = False
        if is_windows():
            SOCKET_CONNECT.close_conn()
        elif is_linux():
            SOCKET_CONNECT._sock.close()
         SOCKET_CONNECT = None
```

Could we set SOCKET_CONNECT = None after closing the connection?

This makes it clear that there is no active connection after closing it and avoids keeping a reference to the closed connection.

saw-jan: yes, we should. 👍

test/gui/environment.py
saw-jan: yes, this is clearly my mistake. I should not move these to here. 👍

Asmitapaudel (approved):

test/gui/helpers/api/webdav_helper.py
```diff
@@ -37,6 +37,7 @@ def resource_exists(user, resource):
 
 def get_file_content(user, resource):
     response = request.get(get_resource_path(user, resource), user=user)
+    assert response.status_code == 200, f"Failed to file content. Status: {response.status_code}"
```
prashant-gurung899:
```suggestion
    assert response.status_code == 200, f"Failed to get file content. Status: {response.status_code}"
```

test/gui/behave.ini
```diff
@@ -12,6 +12,8 @@ logging_level = WARNING
 capture = false
 capture_hooks = false
 show_skipped = false
+; Early fail. Stop the test run on first failure.
+stop = false
```
prashant-gurung899: Confused with comment here.
Won't setting `stop = false` continue the test even after first failure?

test/gui/helpers/SyncHelper.py
```diff
@@ -164,6 +164,19 @@ def clear_socket_messages(resource=''):
         socket_messages.clear()
 
 
+def close_socket_connection():
+    global SOCKET_CONNECT
+    global socket_messages
+    socket_messages.clear()
+    if SOCKET_CONNECT:
+        SOCKET_CONNECT.connected = False
```
anon-pradip: Should not we also remove the GObject watch here? `_connectToSocketServer()` registers a watch with `GObject.io_add_watch(self._sock, IO_IN, self._handle_notify)` but this fn  closes `_sock` without calling `GObject.source_remove(self._watch_id)`. I think this means the event loop will still try to call `_handle_notify` on the next tick, which would attempt `self._sock.recv(1024)` on a closed socket. The `reconnect()` method in `syncstate.py` does call `GObject.source_remove()` so maybe we should do the same here.

test/gui/helpers/SyncHelper.py
```diff
@@ -164,6 +164,19 @@ def clear_socket_messages(resource=''):
...
+        if is_windows():
+            SOCKET_CONNECT.close_conn()
+        elif is_linux():
```
anon-pradip, line 174: I think if `_sock` is `None` here it might cause probelm 🤔 ?. It starts as `None` in `SocketConnect.__init__` and stays `None` if `_connectToSocketServer()` fails. Calling `None.close()` would throw an `AttributeError`. Maybe we need a null check like `_sock is not None` before closing.
`elif is_linux() and SOCKET_CONNECT._sock is not None` ? 🤔

test/gui/helpers/SyncHelper.py
anon-pradip: Not sure if this matters much, but `connected = False` is set before `_sock.close()`. I think it should be the other way around - close first, then update the flag. Otherwise there is a small window where the flag says disconnected but the socket is still open.

test/gui/helpers/SyncHelper.py
```diff
@@ -164,6 +164,19 @@ def clear_socket_messages(resource=''):
         socket_messages.clear()
 
 
+def close_socket_connection():
```
anon-pradip, line 167: So not 100% sure since not much knowdge on socket connection; so feel free to correct me if I am wrong. But maybe something like this would work: Please see below comments for more explanations:

```python
def close_socket_connection():
    global SOCKET_CONNECT
    global socket_messages
    socket_messages.clear()
    if SOCKET_CONNECT:
        if hasattr(SOCKET_CONNECT, '_watch_id') and SOCKET_CONNECT._watch_id:
            GObject.source_remove(SOCKET_CONNECT._watch_id)
        if is_windows():
            SOCKET_CONNECT.close_conn()
        elif is_linux() and SOCKET_CONNECT._sock is not None:
            SOCKET_CONNECT._sock.close()
        SOCKET_CONNECT.connected = False
        SOCKET_CONNECT = None
```

This would need `from gi.repository import GObject` added in the Linux `else` block.

test/gui/helpers/api/webdav_helper.py
```diff
@@ -75,7 +76,8 @@ def create_folder(user, folder_name):
     url = get_resource_path(user, folder_name)
     response = request.mkcol(url, user=user)
     assert response.status_code == 201, (
-        f'Could not create the folder: {folder_name} for user {user}'
+        f'Could not create the folder: {folder_name} for user {user}.'
+        + f'Status: {response.status_code}'
```
anon-pradip:
```suggestion
        + f' Status: {response.status_code}'
```

test/gui/environment.py
```diff
@@ -56,22 +58,11 @@ def after_scenario(context, scenario):
...
-                elif entry.is_dir():
-                    print("Deleting folder: " + entry.name)
-                    shutil.rmtree(prefix_path_namespace(entry.path))
-            except OSError as e:
-                print(f"Failed to delete '{entry.name}'.\nReason: {e}.")
-    # cleanup paths created outside of the temporary directory during the test
-    cleanup_created_paths()
-    delete_project_spaces()
-    delete_created_users()
+        shutil.rmtree(get_config("clientRootSyncPath"))
```
anon-pradip, line 61: But now we are lacking error handling and `prefix_path_namespace()`.

test/gui/behave.ini
anon-pradip, line 16: I think the comment and value might be contradictory here. The comment says "Stop the test run on first failure" but `stop = false` means "continue running after a failure." Also `stop = false` is the behave default so this line is a no-op. Maybe either change to `stop = true` if the intent is to fail fast, or remove these lines entirely?
https://behave.readthedocs.io/en/stable/behave/#cmdoption-stop

saw-jan: yes, true. not changing the current behavior. added the config so that we know how to enable the early fail.

saw-jan: the comment is the config description not the set behavior. added the description because `stop` can mean anything.

saw-jan: and the comment is the config description not the set behavior. Updated the comment, may be it's better now.

saw-jan: Updated the comment, may be it's better now.

test/gui/environment.py
saw-jan: yeah, should be fine. we just delete the test dir. we might need error handling for all the functions in the teardown but really not needed for now.

test/gui/helpers/SyncHelper.py
saw-jan: good point. need to check. 👍

test/gui/helpers/SyncHelper.py
saw-jan: nice one. 👍

test/gui/helpers/SyncHelper.py
saw-jan: removed this because we are doing `SOCKET_CONNECT = None` at the end

test/gui/helpers/SyncHelper.py
saw-jan: 👍 implemented the suggestions

test/gui/helpers/SetupClientHelper.py
```diff
@@ -91,59 +91,67 @@ def start_client():
...
-        sync_paths.update({username: sync_path})
-
-    settings.endGroup()
-
-    settings.sync()
+        settings.sync()
+    except Exception as e:
+        # Do not create config if there is an error.
+        # If not cleared, the 'sync()' will be called by the destructor of QSettings.
+        settings.clear()
```
anon-pradip, line 151:
```python
if settings:
        settings.clear()
```
and we set `settings = None` jut before `try` statement. This will prevent `NameError` if somehow creation of `settings` using `QSettings` constructor failed

anon-pradip (approved):
LGTM :rocket:

test/gui/helpers/SetupClientHelper.py
saw-jan: I will just move the initialization before try. we do not have to try catch the initialization, that's rare

PR #1067, opened 2026-08-31
Abort a hydration request on failure

### LINKED ISSUE #1060: Cfapi: Strange behaviour on hydration of out of sync file
When a file is renamed on the server and we try to hydrate it before the rename is propagated to the file system, we crash.

TheOneRing:
Fixes: #1060

image


guruz (approved):
probably good

PR #1076, opened 2026-09-03
Update system info string format in theme.cpp

jnweiger:
The version info in the about screen showed:

OS: opensuse-leap-6.4.0-150600.23.103-default (build arch: x86_64, CPU arch: x86_64)

BAD: this is confusing, as opensuse-leap never had a version 6.4.0, (this S.U.S.E. version predates the introduction of leap)

I suggest to change the message to include the QSysInfo::productVersion with the os, and print the kernel version 6.4.0 in an extra entry.


TheOneRing:
thx

PR #1091, opened 2026-09-08
Fix check for OAuth::PromptValuesSupported

TheOneRing:


guruz (approved):
Probably good if it works, maybe @ogoffart has a better idea (without exotic new C++ features)

PR #1110, opened 2026-09-17
tell sarif action the checkout path in the container

flimmy:
since the sarif generator is run in a docker container, the checkout paths are different to the paths on the runner itself
see https://github.com/orgs/community/discussions/160738


TheOneRing (approved):

flimmy:
the only "real" documentation of this:
https://docs.github.com/en/actions/how-tos/manage-runners/self-hosted-runners/customize-containers#example-input-for-prepare_job

